PR #4, opened 2018-01-15
Add possibility to choose config file, implement temporary sessions

xates:
- Added the possibility to choose a configuration file by providing its path when initializing the `Client`, instead of using the previously hard-coded "config.ini" file
- Allowing to set the `session_name` to `None` in order to make it "temporary", so that it won't be saved

### DIFF
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -320,7 +320,7 @@ class Client:
             self.user_id = s["user_id"]
 
     def save_session(self):
-        if session_name is not None:
+        if self.session_name is not None:
             auth_key = base64.b64encode(self.auth_key).decode()
             auth_key = [auth_key[i: i + 43] for i in range(0, len(auth_key), 43)]
 
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -61,9 +61,13 @@ class Client:
     invoke every single Telegram API method available.
 
     Args:
-        session_name (:obj:`str`):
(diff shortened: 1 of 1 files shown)


delivrance:
Hi. I wonder how temporary sessions can be useful, you have to log-in with phone number and code every single time if you don't save them.

pyrogram/client/client.py
```diff
@@ -74,7 +78,8 @@ class Client:
     INVITE_LINK_RE = re.compile(r"^(?:https?:\/\/)?t\.me\/joinchat\/(.+)$")
     DIALOGS_AT_ONCE = 100
 
-    def __init__(self, session_name: str, test_mode: bool = False):
+    def __init__(self, config_file: str = "config.ini", session_name: str = None, test_mode: bool = False):
```
delivrance: Maybe we should keep `session_name` as the first parameter so you don't have to explicitly pass a kwarg every time.

xates:
Hi! For example, when the user is logging in on a public/shared device and it is just a one-time login and he doesn't want to save any data.

pyrogram/client/client.py
xates: That is ok for me. As you like it best.

xates: It just seemed me conceptually right to have the config file first :smile:

LeonardoRick:
Could you guys give an example on how to use this config.ini or another solution to change `.session` files location? Thanks a lot!

PR #8, opened 2018-01-24
Support custom callbacks on Client.authorize()

Lonami:


delivrance:
I'm trying to push some changes to your branch, but I think I've messed up something there

PR #14, opened 2018-02-01
Error [400 PHONE_NUMBER_BANNED]

1pyxa1:
when a phone number is banned by Telegram, Client permanently sends requests because it's an UnknownError and 'def authorize(self)' has the code without 'break' or 'raise':
```            
except Exception as e:
                log.error(e, exc_info=True)
```


delivrance:
Nice, thanks! I've just made the error looking more graceful, there's no need to scary people that much by showing the full traceback.

PR #18, opened 2018-02-07
add auto enabling two-step verification if password is not None

1pyxa1:
hi there!
i think it would be great if the `Client` could auto enable two-step verification if you are creating a new account and specified a password. please note, unfortunately i didn't have a chance to test my commit.


delivrance:
@1pyxa1 Hi. For new ideas and enhancements you are welcome to join the group. We can discuss with other people and hear opinions from them too!

1pyxa1:
@delivrance thanks! i am already there

PR #23, opened 2018-02-20
add add_contacts, delete_contacts, get_contacts

1pyxa1:


pyrogram/client/client.py
```diff
@@ -45,14 +45,17 @@
...
-    InputPeerUser, InputPeerChat, InputPeerChannel
+    InputPeerUser, InputPeerChat, InputPeerChannel,
+    InputPhoneContact
 )
 from pyrogram.crypto import AES
 from pyrogram.session import Auth, Session
 from pyrogram.session.internals import MsgId
 from .input_media import InputMedia
 from .style import Markdown, HTML
+from typing import List, Union
```
delivrance: Unfortunately the `typing` module is only available for Python 3.5+. We should remove this in order to support 3.4

pyrogram/client/client.py
```diff
@@ -2306,3 +2309,100 @@ def download_media(self, message: types.Message, file_name: str = None):
...
+                     last_name: str = None,
+                     input_phone_contact_list: List[InputPhoneContact] = None):
+        if (phone is None or first_name is None) and \
+                input_phone_contact_list is None:
+            log.warning("(phone and first_name) or input_phone_contact_list "
+                        "must be not None")
+            return None
+
+        if phone is not None and first_name is not None:
+            if str(phone)[0] != '+':
```
delivrance: `"+" + number.strip("+")` instead looks cleaner.

pyrogram/client/client.py
```diff
@@ -2306,3 +2309,100 @@ def download_media(self, message: types.Message, file_name: str = None):
...
+        # make sure that we send only InputPhoneContact
+        inner_input_phone_contact_list = []
+        for contact in input_phone_contact_list:
+            if isinstance(contact, InputPhoneContact):
+                inner_input_phone_contact_list.append(contact)
+
+        imported_contacts = self.send(
+            functions.contacts.ImportContacts(inner_input_phone_contact_list))
+
+        for user in imported_contacts.users:
```
delivrance: This loop can be replaced by a call to `fetch_peers` which takes care of this

pyrogram/client/client.py
```diff
@@ -2306,3 +2309,100 @@ def download_media(self, message: types.Message, file_name: str = None):
...
+                )
+
+                self.peers_by_id[user.id] = input_peer
+
+                if user.username is not None:
+                    self.peers_by_username[user.username] = input_peer
+
+        return imported_contacts
+
+    def delete_contacts(self, _id: int = None,
```
delivrance: I prefer to keep just one argument (the list), like the raw function does. To add/delete a single contact you simply pass a list with a single element, e.g. [123456789]

pyrogram/client/client.py
```diff
@@ -2306,3 +2309,100 @@ def download_media(self, message: types.Message, file_name: str = None):
...
+                        ids_list: Union[
+                            List[int], List[InputUser]] = None):
+        if _id is None and ids_list is None:
+            log.warning('id or ids_list must be not None')
+            return False
+
+        contacts = self.get_contacts()
```
delivrance: There's no need to check whether the Contact we want to remove exists in the user's contacts list

pyrogram/client/client.py
```diff
@@ -2306,3 +2309,100 @@ def download_media(self, message: types.Message, file_name: str = None):
...
+                            user_id=user.id,
+                            access_hash=user.access_hash)
+                        break
+            inner_ids_list.append(input_user)
+
+        res = self.send(functions.contacts.DeleteContacts(inner_ids_list))
+        return res
+
+    def get_contacts(self, _hash: int = 0):
+        return self.send(functions.contacts.GetContacts(_hash))
```
delivrance: Another call to `fetch_peers` can be made here before returning.

delivrance: Another call to `fetch_peers` can be made here before returning.

delivrance:
Also, since we are dealing with a list of `InputPhoneContact` I'm thinking about adding a new object for this, just like I did for the Album's [InputMedia](https://github.com/pyrogram/pyrogram/blob/master/pyrogram/client/input_media.py). We are writing convenience methods around the raw functions, and I'd like to have convenience types as well to be accessible by using something like `from pyrogram import InputPhoneContact` and not `from pyrogam.api.types import InputPhoneContact`

pyrogram/client/client.py
1pyxa1: what if you don't have a conversation with a contact you want to remove, you need to get ```access_hash```

delivrance: if a call to resolve_peer returns nothing we simply ignore that

delivrance: so, in order to make it working for contacts, the lib must also fetch contacts when it starts, just like get_dialogs

1pyxa1: so, if you don't have a conversation with a contact you cannot remove it? or i didn't get it

pyrogram/client/client.py
1pyxa1: well, it's up to you of course

delivrance: It's also about consistency, other methods like forward_messages, and delete_messages work like this

pyrogram/client/client.py
delivrance: There's no need to have a conversation with a contact in order to remove it.
Here's why: the raw function GetContacts returns this: https://docs.pyrogram.ml/types/contacts/Contacts, and the `users` lists contains the detailed info about all the contacts, with the `access_hash` too.

1pyxa1: yeah, that's why i add ```contacts = self.get_contacts()```, to obtain `access_hash`

delivrance: But getting contacts at the start will enable yet another feature, which is sending messages to contacts by using their phone numbers too

delivrance:
Nice, Telegram is down again

pyrogram/client/client.py
1pyxa1: you mean, you want to add ```self.get_contacts()``` to the ```self.start()```? well, if you accidentally add a contact, you need to perform client restart to be able to delete it, right? or manually get ```access_hash``` from GetContacts

delivrance: Nope, add_contacts will call `fetch_peers` do add the new contact in the internal dictionaries, this is possible because https://docs.pyrogram.ml/types/contacts/ImportedContacts **also** contains the `users` list

1pyxa1: but if you restart the Client, `self.get_dialogs()` won't add new contacts to peers_by_id .._username. i double checked. could you please also test it. maybe my mistake.

1pyxa1: i mean in that case if you haven't start a dialog with new contact

delivrance: Well, of course it won't add new contacts, because the code was missing, I've added it just now. Check new commits on this PR please

1pyxa1: ok, i've understood your idea. thanks!

PR #24, opened 2018-02-21
add get_contacts_hash

1pyxa1:


delivrance:
Also, I don't like having @staticmethods inside the Client class this way. This method and probably some others too will face a refactor and will be moved to something like a utils.py module

1pyxa1:
if you are talking about 
```
def hash20261_32(ints):
    acc = 0
    for id in ints:
            acc = (acc * 20261 * 20261 + id) & 0xffff_ffff
    return acc & 0xffff_ffff
```
or something similar, it was not working for me. i got the algorithm from https://github.com/DrKLO/Telegram

PR #29, opened 2018-02-28
Added sending thumb with video.

23rd:
Added support thumbnail preview in video.
As we can see from official desktop Telegram thumbnails have [size restrictions](https://github.com/telegramdesktop/tdesktop/blob/55e56a6788870e6d67d9938eaabdf87f2ebddb17/Telegram/SourceFiles/storage/localimageloader.cpp#L519) (maximum 90x90 pixels).

### DIFF
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -1346,7 +1346,7 @@ class Client:
         """
         style = self.html if parse_mode.lower() == "html" else self.markdown
         file = self.save_file(video)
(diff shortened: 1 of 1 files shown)


delivrance:
Hi, I think this is something the library itself should take care of by automatically extract a frame from the video and use it as a thumbnail, keeping it as an optional argument is nice, but I'm not sure whether is a good idea to change the method signature just for this

PR #36, opened 2018-03-19
Allow to hook load and save functions to use external session storage…

ezdev128:
PR for the projects like https://github.com/ezdev128/pyrogram-session-redis

### DIFF
--- a/pyrogram/client/__init__.py
+++ b/pyrogram/client/__init__.py
@@ -17,6 +17,6 @@
 # along with Pyrogram.  If not, see <http://www.gnu.org/licenses/>.
 
 from .chat_action import ChatAction
-from .client import Client, Proxy
+from .client import Client, ApiKey, Proxy
 from .parse_mode import ParseMode
 from .emoji import Emoji
--- a/pyrogram/client/__init__.py
+++ b/pyrogram/client/__init__.py
@@ -17,6 +17,6 @@
 # along with Pyrogram.  If not, see <http://www.gnu.org/licenses/>.
 
 from .chat_action import ChatAction
-from .client import Client
+from .client import Client, Proxy
 from .parse_mode import ParseMode
 from .emoji import Emoji
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -175,6 +175,9 @@ class Client:
 
         self.download_queue = Queue()
 
+        self.load_session_hook = None
+        self.save_session_hook = None
+
     def start(self):
         """Use this method to start the Client after creating it.
         Requires no parameters.
@@ -183,7 +186,10 @@ class Client:
             :class:`pyrogram.Error`
         """
         self.load_config()
-        self.load_session(self.session_name)
+        if callable(self.load_session_hook):
+            self.load_session_hook(self)
+        else:
+            self.load_session(self.session_name)
 
         self.session = Session(
             self.dc_id,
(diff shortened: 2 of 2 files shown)


delivrance:
Thanks for this, but Pyrogram is still pretty young and doesn't even persistently cache peers yet, as it should. A redis backend will surely make more sense if there were thousands of peers to store.

pyrogram/client/client.py
```diff
@@ -183,16 +186,18 @@ def start(self):
...
-            self.test_mode,
-            self.proxy,
-            self.auth_key,
-            self.api_key.api_id,
-            client=self
-        )
+        if callable(self.load_session_hook):
+            self.load_session_hook(self)
+        else:
+            self.load_session(self.session_name)
```
delivrance: Not sure if you indented it wrongly, but this piece of code is always needed, here is where Pyrogram creates a new connection to the server.

pyrogram/client/__init__.py
```diff
@@ -17,6 +17,6 @@
 # along with Pyrogram.  If not, see <http://www.gnu.org/licenses/>.
 
 from .chat_action import ChatAction
-from .client import Client
+from .client import Client, ApiKey, Proxy
```
delivrance: I saw your Pyroredis repo, and there is no need to store the proxy settings in a session.
A session should not depend on these information, users should be free to change their proxies regardless of the session used. The same applies for the api key (although I didn't see you using it)

ezdev128: The proxy settings is not stored into the redis session. Please re-check the code again.

delivrance: I thought you were only dealing with sessions, and can't understand why you need Proxy (and ApiKey) to be available there. There is a different method that takes care of loading proxy and api key settings (load_config) and has nothing to do with load_session.

Maybe the auth key creation process inside load_session was confusing you a bit, if that's so we should find a nice way to move it from there. Creating a Session (and auth keys) should only be Client's job, and loading a session must happen before that.

delivrance: What I'm thinking is returning False from load_session in case a session does not exists and the Client needs to create an auth key, this way we can move `self.auth_key = Auth(...)` outside.

ezdev128: During the modification I had several objectives:
1. External **shared** session storage in case of working > 1 of clients (like bot-mode) with the same account credentials.
2. I had a problem with modification session prior to start() because start() overwrites self.session with own self.session = Session(..). So I had to hook it to modify session before self.session.start().

I think will be better if those two actions (session construction and session start) will be in separate methods, like:
```python

	def create_session(self):
		self.session = Session(
			self.dc_id,
			self.test_mode,
			self.proxy,
			self.auth_key,
			self.api_key.api_id,
			client=self
		)

	def start(self):
		self.load_config()
		self.load_session(self.session_name)
		if not self.session:
			self.create_session()
		...
		self.session.start()
		...
```
(or even move self.load_session() under create_session()), so it is possible to do something like:
```python
client = Client(session_name='session_name', api_key=(api_id, api_hash))
client.create_session()
client.session.APP_VERSION = "PyRoGrAm"
...
client.start()

```

delivrance:
Closing because this needs a revamp since the introduction of [storage engines](https://docs.pyrogram.org/topics/storage-engines).

PR #37, opened 2018-03-20
More flexible media downloads with client.download_(media/photo) and client.get_file

EriHoss:
These changes would allow **client.get_file** to write to a given file name or file like object directly
via a new **file_out** parameter.

**file_out** can be a string or file/file like object that implements at least **.write**.

The parameter is optional, and if not specified the old behavior of creating a file
in the current working directory with a generated **.temp** name is kept.

---

**client.download_media** is reworked to allow downloading files to specific locations instead
of only the downloads folder.  The function can still generate a file name for you the way it did prior if desired.

**client.download_media** would no longer make use of a temporary file, as it uses **client.get_files**
new ability to write directly to a file object.

the change in the behavior of **file_name** is breaking for things expecting files to always end up in the "downloads" folder under the current working directory.  And also for things expecting the **block** parameter to be next after **file_name**.

Below is a short test which gives an overview of how the usage would change, and some examples of how **client.download_media** could be used with the changes.

To try this code you can install the version over at:

```pip install git+https://github.com/EriHoss/pyrogram@flexible_media_downloads --upgrade```

```python
from pyrogram import Client
from pyrogram.api import types

# This test assumes that the bot user is sending a jpg photo, or jpg file attachment.
# It uses the sent photo or file in the message update to test the new behavior.

def update_handler(client, update, users, chats):

    if isinstance(update, (types.UpdateNewChannelMessage, types.UpdateNewMessage)):
        message = update.message
        
        if isinstance(message, types.Message) and client.get_me().user.id == message.from_id:
            
            # file_dir has also been added to "client.download_photo", which inherits the
            # behavior of "client.download_media"
            
            # download directly to "direct/test.jpg"
            # directory trees used in file_name are created for you if they do not exist
            client.download_media(message, file_name='direct/test.jpg')
            
            # download to "downloads" in the current directory (the default).
            # folder is created if it does not exist.
            # file has a generated name.
            client.download_media(message)

            # download to the current working directory with a generated name
            client.download_media(message, file_dir='.')
            
            # download to 'custom_dir/custom_dir2' in the current directory.
            # folder tree is created if it does not exist.
            # file has a generated name.
            client.download_media(message, file_dir='custom_dir/custom_dir2')
            
            try:
                # The two arguments cannot be used simultaneously
                client.download_media(message, file_name='test2', file_dir='bad')
            except ValueError:
                print('file_name and file_dir are mutually exclusive')

            # this works because "client.get_file" can now write to file objects, 
            # and "client.download_media" accounts for this.
            # This includes file like objects that only implement .write
            # "client.download_media" will not close the file for you, you must close
            # it since your in control of the file object
            with open('file_object.jpg', 'wb') as f:
                client.download_media(message, file_name=f)
            

            # Test writing to "fake" file objects

            class DummyFile:
                def __init__(self):
                    self._write_cnt = 0

                def write(self, data):
                    # Required
                    self._write_cnt += 1

                def flush(self):
                    # Optional, will be called if it exists
                    print("Write count {}".format(self._write_cnt))

            client.download_media(message, file_name=DummyFile())
            
            # Check if fileno gets used when implemented

            class WrappedFile:
                def __init__(self, filename):
                    self._f = open(filename, 'wb')

                def write(self, data):
                    # Required
                    self._f.write(data)

                def flush(self):
                    # Optional, will be called if it exists
                    self._f.flush()

                def fileno(self):
                   print('fileno called!')
                   return self._f.fileno()
                
                def __enter__(self):
                    return self
                    
                def __exit__(self, exception_type, exception_value, traceback):
                    self._f.close()
            
            with WrappedFile('wrapped_file.jpg') as f:
                client.download_media(message, file_name=f)


def main():
    client = Client("example")
    client.set_update_handler(update_handler)
    client.start()
    client.idle()


if __name__ == "__main__":
    main()
```

### DIFF
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -512,6 +512,10 @@ class Client:
             try:
                 media, file_dir, file_name, done, progress, path = media
 
+                if file_dir is not None:
+                    # Make file_dir if it was specified
+                    os.makedirs(file_dir, exist_ok=True)
+
                 if isinstance(media, types.MessageMediaDocument):
                     document = media.document
 
@@ -2620,7 +2624,8 @@ class Client:
                        file_name: str = None,
                        file_dir: str = 'downloads',
                        block: bool = True,
-                       progress: callable = None):
+                       progress: callable = None
+                       ):
         """Use this method to download the media from a Message.
 
         Args:
@@ -2636,7 +2641,8 @@ class Client:
             file_dir (:obj:`str`, optional):
                 Specify a directory to place the file in if no *file_name* is specified.
                 If *file_dir* is *None*, the current working directory is used. The default
-                value is the "downloads" folder in the current working directory.
+                value is the "downloads" folder in the current working directory.  The
+                directory tree will be created if it does not exist.
 
             block (:obj:`bool`, optional):
                 Blocks the code execution until the file has been downloaded.
@@ -2658,6 +2664,7 @@ class Client:
 
         Raises:
             :class:`pyrogram.Error`
+            :class:`ValueError` if both file_name and file_dir are specified.
         """
 
         if file_name is not None and file_dir is not None:
@@ -2685,6 +2692,7 @@ class Client:
     def download_photo(self,
                        photo: types.Photo or types.UserProfilePhoto or types.ChatPhoto,
                        file_name: str = None,
+                       file_dir: str = None,
                        block: bool = True):
         """Use this method to download a photo not contained inside a Message.
         For example, a photo of a User or a Chat/Channel.
@@ -2696,7 +2704,16 @@ class Client:
                 The photo object.
 
             file_name (:obj:`str`, optional):
-                Specify a custom *file_name* to be used.
+                Specify a custom *file_name* to be used instead of the one provided by Telegram.
+                This parameter is expected to be a full file path to the location you want the
+                photo to be placed. If not specified, the photo will be put into the directory
+                specified by *file_dir* with a generated name.
+
+            file_dir (:obj:`str`, optional):
+                Specify a directory to place the photo in if no *file_name* is specified.
+                If *file_dir* is *None*, the current working directory is used. The default
+                value is the "downloads" folder in the current working directory.  The
+                directory tree will be created if it does not exist.
 
             block (:obj:`bool`, optional):
                 Blocks the code execution until the photo has been downloaded.
@@ -2722,7 +2739,7 @@ class Client:
                 )]
             )
 
-        return self.download_media(photo, file_name, block)
+        return self.download_media(photo, file_name, file_dir, block)
 
     def add_contacts(self, contacts: list):
         """Use this method to add contacts to your Telegram address book.
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -55,6 +55,8 @@ from pyrogram.session.internals import MsgId
 from .input_media import InputMedia
 from .style import Markdown, HTML
 
+from typing import Any
+
 log = logging.getLogger(__name__)
 
 ApiKey = namedtuple("ApiKey", ["api_id", "api_hash"])
@@ -509,7 +511,6 @@ class Client:
 
             try:
                 media, file_name, done, progress, path = media
-                tmp_file_name = None
 
                 if isinstance(media, types.MessageMediaDocument):
                     document = media.document
@@ -535,13 +536,14 @@ class Client:
                                 elif isinstance(i, types.DocumentAttributeAnimated):
                                     file_name = file_name.replace("doc", "gif")
 
-                        tmp_file_name = self.get_file(
+                        self.get_file(
                             dc_id=document.dc_id,
                             id=document.id,
                             access_hash=document.access_hash,
                             version=document.version,
                             size=document.size,
-                            progress=progress
+                            progress=progress,
+                            file_out=file_name
                         )
                 elif isinstance(media, (types.MessageMediaPhoto, types.Photo)):
                     if isinstance(media, types.MessageMediaPhoto):
@@ -558,36 +560,22 @@ class Client:
 
                         photo_loc = photo.sizes[-1].location
 
-                        tmp_file_name = self.get_file(
+                        self.get_file(
                             dc_id=photo_loc.dc_id,
                             volume_id=photo_loc.volume_id,
                             local_id=photo_loc.local_id,
                             secret=photo_loc.secret,
                             size=photo.sizes[-1].size,
-                            progress=progress
+                            progress=progress,
+                            file_out=file_name
                         )
 
                 if file_name is not None:
-                    path[0] = "downloads/{}".format(file_name)
-
-                try:
-                    os.remove("downloads/{}".format(file_name))
-                except OSError:
-                    pass
-                finally:
-                    try:
-                        os.renames("{}".format(tmp_file_name), "downloads/{}".format(file_name))
-                    except OSError:
-                        pass
+                    path[0] = file_name
             except Exception as e:
                 log.error(e, exc_info=True)
             finally:
                 done.set()
-
-                try:
-                    os.remove("{}".format(tmp_file_name))
-                except OSError:
-                    pass
 
         log.debug("{} stopped".format(name))
 
@@ -2177,7 +2165,9 @@ class Client:
                  secret: int = None,
                  version: int = 0,
                  size: int = None,
-                 progress: callable = None) -> str:
+                 progress: callable = None,
+                 file_out: Any = None) -> str:
+
         if dc_id != self.dc_id:
             exported_auth = self.send(
                 functions.auth.ExportAuthorization(
@@ -2225,9 +2215,12 @@ class Client:
                 version=version
             )
 
-        file_name = "download_{}.temp".format(MsgId())
         limit = 1024 * 1024
         offset = 0
+
+        # file object being written
+        f = None
+        close_file, call_flush, call_fsync = False, False, False
 
         try:
             r = session.send(
@@ -2238,30 +2231,49 @@ class Client:
                 )
             )
 
+            if file_out is None:
+                f = open("download_{}.temp".format(MsgId(), 'wb'))
+                close_file = True
+
+            elif isinstance(file_out, str):
+                f = open(file_out, 'wb')
+            elif hasattr(file_out, 'write'):
+                f = file_out
+
+                if hasattr(file_out, 'flush'):
+                    call_flush = True
+                if hasattr(file_out, 'fileno'):
+                    call_fsync = True
+            else:
+                raise ValueError('file_out argument of client.get_file must at least implement a write method if not a '
+                                 'string.')
+
             if isinstance(r, types.upload.File):
-                with open(file_name, "wb") as f:
-                    while True:
-                        chunk = r.bytes
+                while True:
+                    chunk = r.bytes
 
-                        if not chunk:
-                            break
+                    if not chunk:
+                        break
 
-                        f.write(chunk)
+                    f.write(chunk)
+
+                    if call_flush:
                         f.flush()
+                    if call_fsync:
                         os.fsync(f.fileno())
 
-                        offset += limit
+                    offset += limit
 
-                        if progress:
-                            progress(min(offset, size), size)
+                    if progress:
+                        progress(min(offset, size), size)
 
-                        r = session.send(
-                            functions.upload.GetFile(
-                                location=location,
-                                offset=offset,
-                                limit=limit
-                            )
+                    r = session.send(
+                        functions.upload.GetFile(
+                            location=location,
+                            offset=offset,
+                            limit=limit
                         )
+                    )
 
             if isinstance(r, types.upload.FileCdnRedirect):
                 cdn_session = Session(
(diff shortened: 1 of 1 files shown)


delivrance:
Hi! Thanks for this. Let's see...

> These changes would allow client.get_file to write to a given file name or file like object directly
via a new file_out parameter.

I'd rather keep **get_file** untouched, any modification should only improve it and not change its functionality (I'm open to keep discussing about this): the method takes care of downloading and writing bytes and should always write to a temporary file because in case something goes wrong a user won't *ever* see it and won't think is ok where it might be not (take Chrome for example, it creates a temporary file first and renames it later if download was successful).

In other words, if you see your file somewhere you must be 100% sure is correctly downloaded. Also I can't see any advantage of *writing* a file using a file object, a file path should be enough, am I missing something? Is there any notable example of any other python lib that behaves like that?

---

> client.download_media is reworked to allow downloading files to specific locations instead
of only the downloads folder. The function can still generate a file name for you the way it did prior if desired.

I agree in adding support for a custom download location, but I'd like it to be as simple as possible and two mutual exclusive parameters sound a bit convoluted to me. Right now I can think about using a single parameter for either a directory path with a trailing `/` or a file path, without the `/`; this way we can tell custom and generated file_names apart. Any other ideas are welcome.

pyrogram/client/client.py
```diff
@@ -55,6 +55,8 @@
 from .input_media import InputMedia
 from .style import Markdown, HTML
 
+from typing import Any
```
delivrance: Unfortunately, this is Python 3.5+ only. We should remove it for now.

EriHoss:
The idea to allow a file object for the download helper methods arised from the somewhat equivalent methods of the telethon library.

http://telethon.readthedocs.io/en/latest/telethon.html#telethon.telegram_client.TelegramClient.download_media

http://telethon.readthedocs.io/en/latest/telethon.html#telethon.telegram_client.TelegramClient.download_photo

I was thinking along the lines that the user of the library should be in charge of managing the use of a temporary file + swap on complete.  However since pyrograms download functions are pushing work to a background thread for you, I see how it could be impossible to know if an exception occurred which caused the file to be corrupted, making the temp buffer file actually necessary to know if all writes completed.

from what I understand telethons methods run on the current thread, making handling of file exceptions an easier thing.

I would suggest though maybe using an OS allotted temporary file using the tempfile module if **get_file** is still to do the temporary file buffering, or to have the location of the tempfile otherwise configurable.

I could see writing the temp file into the current working directory causing issues if the directory is not writable, or there is something to be overwritten there, etc.

I agree distinguishing directory path vs file path via the trailing path separator would probably be better than adding a new parameter to **download_media** and **download_photo**

I'll have to take a look again later today and see how it could be improved with less breaking changes.

delivrance:
> I was thinking along the lines that the user of the library should be in charge of managing the use of a temporary file + swap on complete.

I don't agree, the lib users only want to easily download their files and not to stress themselves around on details. Definitely lib's job.

> However since pyrograms download functions are pushing work to a background thread for you, I see how it could be impossible to know if an exception occurred which caused the file to be corrupted, making the temp buffer file actually necessary to know if all writes completed.

This is not correct, that's why the above is even more relevant. From the download worker point of view everything happens in the same thread.

Explanation: **get_file** gets actually called from the thread in charge for downloading the file, any exception raised there is caught by the download_worker which will eventually delete the faulty file. When everything's good, get_file will return the relative path of the temporary file to be renamed/moved by download_worker later on, which will then "return" the path to download_media/photo (the lib user).

> I agree distinguishing directory path vs file path via the trailing path separator would probably be better than adding a new parameter to download_media and download_photo.
I'll have to take a look again later today and see how it could be improved with less breaking changes.

Breaking changes are ok if they bring easiness of use (important) or more functionalities. I don't mind changing things if they become cleaner and easier to use, but adding an extra parameter where it can probably be avoided is not something I'm keen on doing.

EriHoss:
One thing I missed here,  **get_file** could possibly allow direct writes to a file-path/file-object without a temp file as optional functionality if it also propagated IO exceptions while doing so.  That would allow you to tell if something went wrong when writing the file, since **get_file** is running in the the foreground.

And when not writing to a file-path or file-object, return to using a temporary buffer file, which **client.download_media** and **client.download_photo** would use.

delivrance:
Sure, but the non-complete file would be visible, which is something I don't really like. The idea is to keep everything hidden under temporary filenames until we are 100% sure the file is downloaded correctly.

EriHoss:
> However since pyrograms download functions are pushing work to a background thread for you, I see how it could be impossible to know if an exception occurred which caused the file to be corrupted, making the temp buffer file actually necessary to know if all writes completed.

What I meant by this is that work submitted by **client.download_media** is running in the download worker. 

Which means that the caller of **client.download_media** really does not have a way to tell if there was an exception inside the worker that caused a corrupt file.  Which does make the temporary file a good idea in this case, because the worker thread can just be in charge of copying the buffer file to the final file if the download runs all the way through without errors. The user will know it did not finish if the file is not there, and the logging shows exceptions etc..

But since **get_file** runs all in the foreground, having a way to write to your own file object and handle any file exceptions that occur on your own might come in handy for more advanced use cases, and forgo the need to rewrite something equivalent to **get_file** on your own if you want to be able to use it in a more fine grained manner.

delivrance:
> Which means that the caller of client.download_media really does not have a way to tell if there was an exception inside the worker that caused a corrupt file.

You are missing this: there actually is a way, and is the same download_worker has for telling whether get_file failed or not, which is, returning either None or the file path.

EriHoss:
> Explanation: get_file gets actually called from the thread in charge for downloading the file, any exception raised there is caught by the download_worker which will eventually delete the faulty file. When everything's good, get_file will return the relative path of the temporary file to be renamed/moved by download_worker later on, which will then "return" the path to download_media/photo (the lib user).

This does not seem to be the case currently due to the broad exception clauses inside **get_file**.

**download_media** cannot catch file exceptions from it because it does not seem that they propagate from the broad exception clause starting on line 2232.

> You are missing this: there actually is a way, and is the same download_worker has for telling whether get_file failed or not, which is, returning either None or the file path.

Yes I did miss this, I see now that *download_media* is returning path[0] which is set in the worker.

delivrance:
> This does not seem to be the case currently due to the broad exception clauses inside get_file.
download_media cannot catch file exceptions from it because it does not seem that they propagate from the broad exception clause starting on line 2232.

You are right, download_media isn't catching any exception raised by get_file and it is relying on its return value, which will be None if something went bad. And this is actually a bug, since download_media doesn't check whether get_file failed or not

EriHoss:
> Also, I still don't get why is passing a file object any easier than passing a file path when writing, it makes more sense when reading a file.

It could do without it if you wanted to keep **get_file** purely high level. 

I think mostly, having the temp file be written to a more discrete location would be good though.

along with maybe having **download_media** able to write to a specific location.

EriHoss:
get_file's only change from master is that it now returns a path to an OS allotted temporary file.

I'm unsure if the **if** branch at:

https://github.com/pyrogram/pyrogram/blob/aa8125d7a2c25095b8dbd0159f0a982c6f22fd31/pyrogram/client/client.py#L2269

Might actually need to be **elif** however, I'm unfamiliar with how **types.upload.FileCdnRedirect** works, but it looks a little strange.

This includes the fix you just committed.

Latest commits testing something along the lines of this.

```python
from pyrogram import Client
from pyrogram.api import types

def update_handler(client, update, users, chats):
    # This test assumes that the bot user is sending a jpg photo, or jpg file attachment

    if isinstance(update, (types.UpdateNewChannelMessage, types.UpdateNewMessage)):
        message = update.message
        
        if isinstance(message, types.Message) and client.get_me().user.id == message.from_id:
            
            # file with generated name in the "directory/" folder under the current directory
            # directory is created if it does not exist
            client.download_media(message, file_name='directory/')
            
            # two directories down, with generated name
            # directories are created if they do not exist
            client.download_media(message, file_name='directory/directory2/')
            
            # file named 'file.jpg' in the downloads folder, under the current directory
            client.download_media(message, file_name='file.jpg')
            
            # use a path separator elsewhere (not at the end) to indicate a specific path
            # downloads to CWD/file
            client.download_media(message, file_name='./file.jpg')
            
            # download to CWD/direct/file.jpg
            # the directory tree is created if it does not exist
            client.download_media(message, file_name='./direct/file.jpg')


def main():
    client = Client("example")
    client.set_update_handler(update_handler)
    client.start()
    client.idle()


if __name__ == "__main__":
    main()
```

delivrance:
Good, I'll review it soon.

> I'm unsure if the if branch at: ... Might actually need to be elif however, I'm unfamiliar with how types.upload.FileCdnRedirect works, but it looks a little strange.

Logically irrilevant in this case (and probably confusing), but yeah, elif would be more readable and save an useless check as well. Either this or `if not chunk: ...` should return. The former is better.

pyrogram/client/client.py
```diff
@@ -2279,7 +2301,18 @@ def get_file(self,
                 cdn_session.start()
 
                 try:
-                    with open(file_name, "wb") as f:
+                    # cant fdopen the closed file descriptor which could be closed due
```
delivrance: Code flow will follow either one of these two branches, doing so there will be a zombie fd in case a file is stored in CDNs.

delivrance: I actually tested your code above before reading you were actually removing it, but I saw some extra unused files in the temp directory (Windows) and that's weird

delivrance: Ok, the reason is that you did not close the fd, and we get an OSError which will be swallowed by the try catch below.

EriHoss: Yes, this is what I was thinking when you mentioned that only one of those branches is going to be hit.

The first call to **tempfile.mkstemp** here:

https://github.com/EriHoss/pyrogram/blob/8796e857af44918a05cf37fb3b5f42cbca6fb7c4/pyrogram/client/client.py#L2264

Can be moved into the first branch above where the file descriptor is opened in the **with** statement to cure this.

And this logic can be removed:

https://github.com/EriHoss/pyrogram/blob/8796e857af44918a05cf37fb3b5f42cbca6fb7c4/pyrogram/client/client.py#L2308

I'll go ahead and change this.

pyrogram/client/client.py
```diff
@@ -571,26 +584,33 @@ def download_worker(self):
...
             finally:
                 done.set()
 
                 try:
-                    os.remove("{}".format(tmp_file_name))
-                except OSError:
-                    pass
+                    os.remove(tmp_file_name)
+                except OSError as e:
+                    if not e.errno == errno.ENOENT:
```
delivrance: Care to explain why is this necessary?

EriHoss: The ``if not e.errno == errno.ENOENT`` is for logging an exception in the case that the file cannot be removed, but it is for some other reason aside from that it does not exist. Such as if it is still in use (at least on windows), which might be helpful in revealing a subtle bug if there is one.

It would be better off as ``if e.errno != errno.ENOENT`` actually, not sure why I wrote it with the not.

This could be worth doing elsewhere possibly.

delivrance: Using tempfile just made this whole thing overcomplicated in my opinion, now we need to handle file descriptors manually outside a context manager. a .temp file in the current working directory was doing no harm, we could also have make it hidden by prepending a dot eventually.

delivrance: The method was already generating unique file names, so I can't really see any benefit in using tempfile here

EriHoss: > Using tempfile just made this whole thing overcomplicated in my opinion, now we need to handle file descriptors manually outside a context manager. a .temp file in the current working directory was doing no harm, we could also have made it hidden by prepending a dot eventually.

The FD should be closed inside **get_file** leaving just the unopened temp file in an OS specific location.  Users of **get_file** should never need to manually close an FD.

The "." is not what is causing the **file_name** parameter to be treated as a full path.  It is the presence of either "/" or "\" in the path, and no trailing directory separator, which causes it to treat it as a full file path.

using a **file_name** like this ".filename" should result in a hidden file being placed in the "downloads" folder of the current directory.  The "." is not interpreted as anything special.

EriHoss: The idea of a OS provided tempfile better guarantees that the file can actually be created.

For instance, if I were to decide to run the client on windows out of the programs folder directory with the working directory where the script is, or otherwise run the client in a non writable working directory, than the temp file could not be written there.

Same goes for the default downloads folder, which is always getting created in the working directory.

EriHoss: > Using tempfile just made this whole thing overcomplicated in my opinion, now we need to handle file descriptors manually outside a context manager. a .temp file in the current working directory was doing no harm, we could also have made it hidden by prepending a dot eventually.

I slightly misinterpreted what you were saying about having the FD outside of a with statement, above.  Yes I agree that is a bit precarious. 

Though the code could make use of:

```tempfile.NamedTemporaryFile('wb', delete=False)`` inside of a **with** context manager instead and be equivalent.

**Comment by EriHoss** [comment]
https://github.com/pyrogram/pyrogram/pull/37/commits/0f4e29584ac46af38e8605f9720e3b0e8189693d should remove the need to deal with the file descriptor manually in **get_file**, using an equivalent use of **tempfile.NamedTemporaryFile**

**Comment by delivrance** [comment]
Wanna move the conversation to Telegram? @PyrogramChat
We can more easily discuss there, if you like.

**Comment by EriHoss** [comment]
I can pop on there sometime tommorow if that is easier.

**Comment by delivrance** [comment]
I've cleaned it a bit. PR is ready to merge, but feel free to ping me if there's anything else

File: pyrogram/client/client.py [hunk 4]
```diff
@@ -580,37 +572,29 @@ def download_worker(self):
...
-                            os.makedirs(os.path.dirname(file_name), exist_ok=True)
+                        os.remove(final_file_path)
+                    except OSError:
+                        pass
 
-                        # avoid errors moving between drives/partitions etc.
-                        shutil.move(tmp_file_name, os.path.join(download_directory, file_name))
-                    except OSError as e:
-                        log.error(e, exc_info=True)
+                    os.renames(temp_file_path, final_file_path)
```
EriHoss: Just one thing with this.

In my case, my projects live on another disk drive. This lead me to discover that **os.renames** will error out if you try to rename to a location on another drive.

Since on Windows where I am currently testing things, the temp directory always exists on **C:**, unless you have 

When it goes to move the temporary file from **C:** to **J:** where the tests working directory is, it will be unable to do this and cause an exception.

Hence the use of **os.makedirs** and **shutil.move** in my code instead.  

**shutil.move** is able to move files from one disk / partition to another without this error.

https://docs.python.org/3/library/os.html#os.rename

According to the **os.rename** doc, this can also be the case on some Unix systems.

https://docs.python.org/3/library/shutil.html#shutil.move

**shutil.move** uses **os.rename** if **dst** and **src** are on the same disk, but you must make the directory tree that the file is going to be moved to before you move it.

EriHoss: This sort of error occurs running the second test snippet with the use of **os.rename**, when the temp file exists on a different disk that the destination.

https://i.imgur.com/8Pg91JL.png

EriHoss: Something along the lines of this at the **if** branch on line 575 fixes for me:

``` python
if temp_file_path:
    final_file_path = os.path.join(directory, file_name)

    final_file_dir = os.path.dirname(final_file_path)

    if final_file_dir:
        os.makedirs(final_file_dir, exist_ok=True)

    shutil.move(temp_file_path, final_file_path)
```

**shutil.move** overwrites **final_file_path**, so there should not be a need to delete any existing file first.

This should work unless there is something about **shutil.move** that you want to avoid that I am missing.

delivrance: Very good, but this would work the same:

``` python
if temp_file_path:
    final_file_path = os.path.join(directory, file_name)
    os.makedirs(directory, exist_ok=True)
    shutil.move(temp_file_path, final_file_path)
```

The directory path is already computed.

A thing we might also want to improve is the path looking, 'cause I'm getting paths like `asd/asddd\sticker.webp`

EriHoss: I am unsure where the mixed path separator may be coming from at the moment.

But as long as **directory** is never an empty string the above should do.

Because ```os.makedirs("", exist_ok=True)``` with throw **FileNotFoundError**.

EriHoss: > A thing we might also want to improve is the path looking, 'cause I'm getting paths like asd/asddd\sticker.webp

What could be done to make the returned path consistent is call **os.path.abspath** on it when putting the result in **path[0]**.  That way it always has the correct path separators for your platform.

delivrance: > What could be done to make the returned path consistent is call os.path.abspath on it when putting the result in path[0]. That way it always has the correct path separators for your platform.

Good, an absolute path looks even better than a relative one, but we need both **os.path.abspath** and the substitution because I'm getting `/home/dan/asd/asdads\asdafnaopinf\\\\asfopo/aosfnia\\aasdd` in Linux with os.path.abspath only.

> But as long as directory is never an empty string the above should do.

There's a check, if the directory is empty it will default to "downloads", so yeah, it can never be empty.

pyrogram/client/client.py
```diff
@@ -28,6 +28,8 @@
 import tempfile
 import threading
 import time
+import shutil
```
delivrance: Right, pycharm is likely to move it somewhere else anyway ehe

EriHoss:
Paths currently look good on my end with this change.

https://i.imgur.com/gOFSToA.png

Also running this change in my client project currently with no issues so far.

https://github.com/EriHoss/TGMiner/commit/7cad713651f2e12d65e8906a0b3c39a1e44d70f8

delivrance:
Same for me on Windows, is a Linux only issue apparently

Issue #68, opened 2018-04-30
Session threads left running after Client.stop, occurs in long running process


EriHoss:
Threads created in **Session.start()** appear to be left running in a loop after stopping a client that has been running for some time.

On CTRL-C this causes the program to hang, because the threads are not deamon threads and continue to keep the program alive.

It appears that new threads might be getting created before the previous successfully exited when session reconnects while running?

In my project I have added something like this for debugging:

```python
client.start()
client.idle()
time.sleep(5)
for i in (x for x in threading.enumerate() if x.name != 'MainThread'):
    print(i.name)
```

Occasionally after a long run I get this output on CTRL-C (put some debug prints in **Syncer.remove** because I thought it might be a deadlock there but it is not):

```python
Syncer.remove lock before
Syncer.remove cls.sync(client) after
Syncer.remove cls.stop() before
Syncer.stop cls.event.set() before
Syncer.stop cls.event.set() after
Syncer.remove cls.stop() after
Syncer.remove lock after
NetWorker#1
RecvThread
NextSaltThread
PingThread

... program hangs ...
```

I think the problem might be in **Session.start()** itself but I am not certain.

EriHoss:
This is the bulk of what my program is doing (cut down) for testing purposes.

I am experiencing this on windows.

```python
import threading
import time

from pyrogram import Client, RawUpdateHandler, Message
from pyrogram.api import types as api_types
from pyrogram.client import message_parser

import uuid

app = Client("pyrogram.session", api_id=999999, api_hash="hash_here")


def update_handler(client, update, users: dict, chats: dict):
    if not isinstance(update, (api_types.UpdateNewChannelMessage, api_types.UpdateNewMessage)):
        return

    update_message: api_types.Message = update.message

    parsed_message: Message = message_parser.parse_message(
        client,
        update.message,
        users,
        chats
    )

    if isinstance(update_message.media, (api_types.MessageMediaPhoto, api_types.MessageMediaDocument)):
        m_name = str(uuid.uuid4())
        print('Download media: downloads/'+m_name)
        client.download_media(parsed_message, file_name=m_name, block=False)

    print(update_message.message)


app.add_handler(RawUpdateHandler(update_handler))

app.start()
app.idle()

time.sleep(5)

for i in (x for x in threading.enumerate() if x.name != 'MainThread'):
    print(i.name)

```

delivrance:
Hi, thanks for the detailed report!

[...]

Issue #40, opened 2018-03-22
Better handling of sessions


surkum:
While downloading a great number of media files I'm getting this exception that is nor treated nor raised

```[420 FLOOD_WAIT_X]: A wait of 1369 seconds is required
Traceback (most recent call last):
  File "/home/user/tmp/telegram_backup/lib64/python3.6/site-packages/pyrogram/client/client.py", line 338, in download_worker
    progress=progress
  File "/home/user/tmp/telegram_backup/lib64/python3.6/site-packages/pyrogram/client/client.py", line 2143, in get_file
    dc_id=dc_id
  File "/home/user/tmp/telegram_backup/lib64/python3.6/site-packages/pyrogram/client/client.py", line 554, in send
    r = self.session.send(data)
  File "/home/user/tmp/telegram_backup/lib64/python3.6/site-packages/pyrogram/session/session.py", line 402, in send
    return self._send(data)
  File "/home/user/tmp/telegram_app/lib64/python3.6/site-packages/pyrogram/session/session.py", line 388, in _send
    Error.raise_it(result, type(data))
  File "/home/user/tmp/telegram_backup/lib64/python3.6/site-packages/pyrogram/api/errors/error.py", line 67, in raise_it
    )(x)
pyrogram.api.errors.exceptions.flood_420.FloodWait: [420 FLOOD_WAIT_X]: A wait of 1369 seconds is required
......
``` 

and the exception is reported again and again until flooding wait time pass, the file is downloaded and then it is raised again with every download command.

the code which produces the exception  is: 

`filename = client.download_media(message, block=True, file_name=name)`

delivrance:
What raises the FLOOD_WAIT_X error is ExportAuthorization.
The problem is that Pyrogram creates new keys for every file downloaded, and the server is limiting us in doing so. We should instead create one, cache it and re-use it when needed.

delivrance:
@surkum Can I ask you how many files have you tried to download? More or less. Just to get an idea of when such situation would occur.

surkum:
I was trying to backup a long conversation (14 messages), and all it's media files(~2000), it started to fail around 400

delivrance:
@surkum Hi. I've pushed some changes to a new branch. Use `pip3 install -U git+https://github.com/pyrogram/pyrogram@cache-sessions` to test this out and tell me how it goes.

surkum:
[...]

PR #69, opened 2018-05-01
Make sure started Session is always stopped upon exceptions in Client.get_file

EriHoss:
In reference to: https://github.com/pyrogram/pyrogram/issues/68

I see there is a TODO on this function but here is a fix.


delivrance:
Merging this, but I'm gonna rework this part to accommodate the changes in #40.

PR #72, opened 2018-05-07
Proposal for using abstractmethods and registering virtual classes

JosXa:


pyrogram/client/ext/base_client.py
```diff
@@ -75,15 +76,16 @@ def __init__(self):
         self.download_queue = Queue()
         self.download_workers_list = []
 
-    def send(self, data: Object):
-        pass
+    @abstractmethod
+    def send(self, data: Object): pass
```
delivrance: Function body in the same line? Hm I don't quite like it in python, even if is just `pass`

pyrogram/client/methods/bots/callback_query/answer_callback_query.py
```diff
@@ -20,7 +20,9 @@
 from ....ext import BaseClient
 
 
-class AnswerCallbackQuery(BaseClient):
+# noinspection PyUnresolvedReferences
```
delivrance: I don't really like this either. How can we be 100% sure is pycharm's fault here?

pyrogram/client/methods/contacts/add_contacts.py
```diff
@@ -32,7 +34,7 @@ def add_contacts(self, contacts: list):
             On success, the added contacts are returned.
 
         Raises:
-            :class:`Error <pyrogram.Error>`
+            :lass:`Error <pyrogram.Error>`
```
delivrance: lass

pyrogram/client/ext/base_client.py
JosXa: Pretty sure that's standard for @abstractmethods

pyrogram/client/methods/contacts/add_contacts.py
JosXa: lol

pyrogram/client/methods/bots/callback_query/answer_callback_query.py
JosXa: Official docs use it in exactly this way. Fairly sure it's a Pycharm issue, and it works flawlessly

JosXa: I lied. Official docs only say
"Changed in version 3.3: Returns the registered subclass, to allow usage as a class decorator."
-- https://docs.python.org/3/library/abc.html#abc.ABCMeta.register

Examples on the internet use it like this.

delivrance:
Closing because I think abc is unnecessary boilerplate.
Change my mind.

PR #77, opened 2018-05-26
Manual locking proposal

JosXa:
(untested)


delivrance:
As we discussed already, this won't work ^^

Issue #88, opened 2018-06-18
Update on deleted message


mateusc-ciandt:
Is there a way to receive an update event when a channel message is deleted?

Using the following code I get new messages and edited messages, but when a message is deleted, it does not trigger my method:
```python
@app.on_message(Filters.channel)
def from_channel(client, message):
    print(message)
```

delivrance:
Hi! For completeness I'll answer here too, although the discussion started and is ongoing in the [official group chat](https://t.me/pyrogramchat). ^^

There's no way at the moment other than using `on_raw_update` (aka `RawUpdateHandler`) and manually filtering the updates.

The current idea is to add a whole new handler, namely `DeleteMessagesHandler`, and its decorator `on_deleted_messages` which will register a function with a **list of messages** (maybe [Messages](https://docs.pyrogram.ml/pyrogram/types/Messages)?) in its arguments instead of a **single Message**. This is preferred due to the fact that people can delete up to 100 messages at once and Telegram returns a single update containing the list of the deleted message ids; having something like `on_message(Filters.deleted)` is cool, but the callback function would be called up to 100 times for a single delete Update in order not to break the function signature and its consistency.

mateusc-ciandt:
Hi @delivrance 
I totally agree with you, I'm working on the implementation of this handler. 

I will soon open the pull request! :smiley:

mateusc-ciandt:
Like I said, I'm new to the python world. Feel free to give feedback on my Pull Request @delivrance :smile: 

Thanks

vitalibr:
Like I said, I'm new to the python world. Feel free to give feedback on my Pull Request @delivrance smile

Thanks

tahayasin2:
need just delete message

PR #91, opened 2018-06-19
Add deleted messages handler

vitalibr:
Added a handler for deleted messages according to the open enhancement #88

### DIFF
--- a/pyrogram/client/dispatcher/dispatcher.py
+++ b/pyrogram/client/dispatcher/dispatcher.py
@@ -25,7 +25,7 @@ from threading import Thread
 import pyrogram
 from pyrogram.api import types
 from ..ext import utils
-from ..handlers import RawUpdateHandler, CallbackQueryHandler, MessageHandler
+from ..handlers import RawUpdateHandler, CallbackQueryHandler, MessageHandler, DeletedMessagesHandler
 
 log = logging.getLogger(__name__)
 
@@ -39,6 +39,11 @@ class Dispatcher:
     EDIT_MESSAGE_UPDATES = (
         types.UpdateEditMessage,
         types.UpdateEditChannelMessage
+    )
+
+    DELETE_MESSAGE_UPDATES = (
+        types.UpdateDeleteMessages,
+        types.UpdateDeleteChannelMessages
     )
 
     MESSAGE_UPDATES = NEW_MESSAGE_UPDATES + EDIT_MESSAGE_UPDATES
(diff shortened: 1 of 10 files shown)


pyrogram/client/dispatcher/dispatcher.py
```diff
@@ -161,6 +174,22 @@ def update_worker(self):
                                                  else None)
                         )
                     )
+
+                elif isinstance(update, Dispatcher.DELETE_MESSAGE_UPDATES):
```
delivrance: I was about to suggest to put this branch into the above, but then I realized that `parse_messages` would require some changes to also parse deleted messages. Using another utility function makes sense here. Good job!

pyrogram/client/ext/utils.py
```diff
@@ -711,6 +711,22 @@ def parse_messages(
...
 
 
+def parse_deleted_messages(
+        messages: list,
+        channel_id: int
+) -> pyrogram_types.Messages:
+    parsed_messages = []
+
+    for message in messages:
+        m = pyrogram_types.Message(
```
delivrance: Since `m` is used only once I usually prefer something like: 
``` python
parsed_messages.append(
    pyrogram_types.Message(
        message_id=message,
        chat=(pyrogram_types.Chat(id=channel_id, type="channel")
              if channel_id is not None
              else None)
    )
)
```

What do you think?

pyrogram/client/ext/utils.py
```diff
@@ -711,6 +711,22 @@ def parse_messages(
...
+            message_id=message,
+            chat=(pyrogram_types.Chat(id=channel_id, type="channel") if channel_id is not None else None)
```
delivrance: Here we must use `id=int("-100" + str(channel_id))` to convert the channel_id into its Bot API id representation.

pyrogram/client/filters/filters.py
```diff
@@ -89,13 +89,13 @@ class Filters:
     venue = build("Venue", lambda _, m: bool(m.venue))
     """Filter messages that contain :obj:`Venue <pyrogram.api.types.pyrogram.Venue>` objects."""
 
-    private = build("Private", lambda _, m: bool(m.chat.type == "private"))
+    private = build("Private", lambda _, m: bool(m.chat and m.chat.type == "private"))
```
delivrance: Good, extra checks here won't do any harm. We still need to find a way to make private, group and channel filters work for deleted messages, though

delivrance:
Nice PR. Hardly anything to change except some code style and docstring fixes.
Good job! Thanks

mateusc-ciandt:
:medal_sports:

PR #94, opened 2018-07-01
added reverse option to get_history

YoilyL:


YoilyL:
Can you give me a test example to reproduce where it doesn't work as expected?
I worked fine in all my tests

delivrance:
Simply `app.get_history("me", limit=3, offset=3, reverse=True)`

I would have expected something like: go up 3 messages from the last one, get the 3 messages down starting from there.

delivrance:
Is not a real problem if you take `offset` as a real offset. The problem is how you think the messages should be returned, because there's another logic involved for `reverse=False`, we just need to make it consistent for both cases

YoilyL:
This is the correct behavior I was aiming for. People trying to get from message_id 1 to the last message. 
With your example 
`app.get_history("me", limit=3, offset=3, reverse=True)` I'm trying to get the messages with id [10,11,12]

YoilyL:
The point is that usually people trying to get the history of a chat usually want to go in order from first message to last.
With this method you can simply loop through the chat and get messages from id 1 to the last id.

delivrance:
Another problem is that id 1 is not necessarily the first message of a chat; it works for supegroups and channels, but not for private and basic group chats. In fact, the only way to get the very first message in a private chat is to iterate over the message backwards. We need to make get_history work for both cases

YoilyL:
Oh damn... I only tested on a channel

😙
🙈🙈

delivrance:
I've modified it a bit, but I'm still unsure about `offset` naming and the fact the last message starts at index 0:

- Messages are now always returned from older to newer, regardless of any parameter.
- There are three main offsets: offset (by message index, last message = 0), offset_id (by message id), offset_date (by message date).
- The add_offset parameter is used to move around the main offset (can be negative for newer messages).
- The downwards parameter set to True will make the method pick newer messages (below) instead.

delivrance:
Some usage examples at https://t.me/PyrogramChat/25453 and https://t.me/PyrogramChat/25454

delivrance:
Hi @YoilyL are you still up in reviewing this? I think I'm going to merge it soon, but first we need to think about better parameter names

PR #111, opened 2018-08-30
Allow to set audio thumbnail when using send_audio

zeroone2numeral2:
With the [Bot API 4.0 update](https://core.telegram.org/bots/api#july-26-2018), `sendAudio` allows bots to pass a thumbnail - making it possible for pyrogram to include a `thumb` parameter in its convenience method `send_audio`

### DIFF
--- a/pyrogram/client/methods/messages/send_audio.py
+++ b/pyrogram/client/methods/messages/send_audio.py
@@ -35,6 +35,7 @@ class SendAudio(BaseClient):
                    duration: int = 0,
                    performer: str = None,
                    title: str = None,
+                   thumb: str = None,
                    disable_notification: bool = None,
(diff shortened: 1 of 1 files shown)


pyrogram/client/methods/messages/send_audio.py
```diff
@@ -74,6 +75,11 @@ def send_audio(self,
             title (``str``, *optional*):
                 Track name.
 
+            thumb (``str``, *optional*):
+                Audio thumbnail.
```
delivrance: For such method parameters I usually take docstrings from the bot api docs (when they exist), with some small changes to adapt the texts or fix typos.

delivrance:
Thanks!

Issue #133, opened 2018-10-05
Allow specifying more than one prefix in Filters.command


delivrance:
Just like the `command` argument of the command filter can accept a list of strings, it makes sense to allow specifying multiple prefixes as well, so that `Filters.command(["start", "help"], ["/", "!"])` can accept `/start`, `!start`, `/help` and `!help`.

delivrance:
#137 Merged

PR #137, opened 2018-10-14
Allow multiple command prefixes in a single filter

delivrance:
From #133

**TODO:**

- [x] Update docstrings


delivrance:
From #133

delivrance:
I forgot to push to `develop` before branching. That's why I've got irrelevant commits here.

Issue #143, opened 2018-10-18
Add checks for file sizes when uploading


delivrance:
File size limit is exactly 1500 MiB (1 MiB = 1024 KiB); if a bigger file is attempted to upload, a cryptic RPCError is shown. Let's make use of a nicer error instead.

TolichP:
I am interested to do it, can I?

delivrance:
Sure!

delivrance:
Solved in #146

PR #146, opened 2018-10-19
Add file size limit error

TolichP:
Add file size limit error for #143

### DIFF
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -1123,6 +1123,7 @@ class Client(Methods, BaseClient):
                   progress_args: tuple = ()):
         part_size = 512 * 1024
         file_size = os.path.getsize(path)
+        assert file_size < 1500 * 1024 * 1024, "Couldn't upload file bigger 1500MiB"
         file_total_parts = int(math.ceil(file_size / part_size))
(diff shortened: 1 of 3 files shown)


pyrogram/client/client.py
```diff
@@ -1123,6 +1123,7 @@ def save_file(self,
                   progress_args: tuple = ()):
         part_size = 512 * 1024
         file_size = os.path.getsize(path)
+        assert file_size < 1500 * 1024 * 1024, "Couldn't upload file bigger 1500MiB"
```
delivrance: - Not a fan of `assert`s in these scenarios. Raising `ValueError` is preferred.
- Files of exactly 1500 MiB are supported, so you wanted `<=` instead.
- I also like more descriptive error messages.

Suggestion:

```suggestion
        if file_size > 1500 * 1024 * 1024:
            raise ValueError("Telegram doesn't support uploading files bigger than 1500 MiB")
```

TolichP:
Sorry, I'm new to making pull requests and I have done something strange with commits. I tried to apply the proposed changes.

delivrance:
@TolichP Nope, you just need to add more commits to your forked repo in the `add-filesize-limit-error` branch (which you did already), they will be automatically added here.

PR #147, opened 2018-10-19
Update README.rst

TolichP:
Add information for contributors.

When I tried to launch project locally, I faced with such problem - `client.py` contains `from .all import objects`, but no file `all.py` provided.
Information on the necessity of executing some command could be useful in the `README.rst`


delivrance:
Thanks for the suggestion, but I'm not gonna add it in the readme directly. I prefer having a dedicated contributors' page instead.

Also, I see you wrote `python setup.py generate`, but that does simply nothing because it misses a required argument. The correct command is `python setup.py generate --api`

TolichP:
Ok, thanks for the explanation.

PR #148, opened 2018-10-26
Updated doc info about thumbnail's size.

23rd:
I noticed again that the documentation still says about the 90 pixel limit.
After six months of testing, I can say that the formula `width + height <= 400 pixels` works perfectly for thumbnails.
I don't see any reason to mislead users with wrong limitation.

### DIFF
--- a/pyrogram/client/methods/messages/send_animation.py
+++ b/pyrogram/client/methods/messages/send_animation.py
@@ -75,7 +75,7 @@ class SendAnimation(BaseClient):
             thumb (``str``, *optional*):
                 Thumbnail of the animation file sent.
                 The thumbnail should be in JPEG format and less than 200 KB in size.
-                A thumbnail's width and height should not exceed 90 pixels.
+                A sum of a thumbnail's width and height should not exceed 400 pixels.
                 Thumbnails can't be reused and can be only uploaded as a new file.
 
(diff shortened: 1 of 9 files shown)


delivrance:
Hi, I know you tested a lot on this and I had no reason to edit this info, until the Bot API has updated, so I took the value straight from their docs.

I prefer to mention official and conservative information about limits, when available, instead of relying on amateur tests that push limits to their edges (this is what I call misleading actually, because we can't be 100% sure this will remain true forever; it is not written anywhere).

Also, saying you **should** doesn't mean you cannot try higher values by yourself and use them if you see they work (there's no hard limit in Pyrogram in that sense), it just implies the method guarantees that within said limits you are safe.

tl;dr: We can't really guarantee that the 400px sum limit is not going to be changed in future, better stick with official limits so that people can't blame the library in case something goes wrong.

23rd:
Oh, I thought officially limits are nowhere to be found except in the code of official applications.
Can you share a link?

delivrance:
The update is relatively new indeed. Example: https://core.telegram.org/bots/api#senddocument
Also search for "90" throughout the page

23rd:
Okay, I see. 
Have no any questions.
Thanks.

PR #159, opened 2018-11-14
fix RawUpdateHandler

Jafte:
Hello! 

I have been using RavUpdateNandler for a long time, but after the update it began to behave strangely. As I understood from the code, after processing the raw update from the telegrams, the variable with the raw update is overwritten and there is no longer a "raw" update.

Hotfix working for me.

### DIFF
--- a/pyrogram/client/dispatcher/dispatcher.py
+++ b/pyrogram/client/dispatcher/dispatcher.py
@@ -129,7 +129,7 @@ class Dispatcher:
                 if parser is None:
                     continue
 
-                update, handler_type = parser(update, users, chats)
+                parsed_update, handler_type = parser(update, users, chats)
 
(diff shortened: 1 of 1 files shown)


delivrance:
> the variable with the raw update is overwritten and there is no longer a "raw" update.

@Jafte Indeed! It should not be overwritten. Thank you for the report and the fix.

**Note:** I'm not a fan of raw updates because working with them makes code quite messy. If I can know what do you need exactly from raw updates maybe we can add support and turn it an higher-level update that's easier to use (see Message, CallbackQuery, etc...)

Jafte:
> If I can know what do you need exactly from raw updates

I am using custom Client class without inputs interruption with tornado server for management several accounts through rest api. RawUpdateHandler, first of all, helps me in debug.

Issue #197, opened 2019-01-08
Plugins loading should respect alphabetical order


zeroone2numeral2:
This is somehow related to #183 but it's probably worth to have a different smaller issue.

Right now plugins are loaded "unordered", because [`pathlib.Path.glob()`](https://github.com/pyrogram/pyrogram/blob/4d7b1bafac7904aee8f683fa781651130349b9ea/pyrogram/client/client.py#L1028) makes pyrogram loop through plugins in an undefined order (apparently, the order in which plugins are loaded is the same of `ls -U`'s output).

Since there's no way as of now to define the loading order of our plugins, it would be nice to have them at least loaded alphabetically (which gives a bit of control to the user, who can name them according to his preference). Should be easier and quickier to pick than implementing #183's import system, as a temporary solution

Issue #183, opened 2018-12-24
Allow specifying which plugins to load and in which order


delivrance:
Currently, the plugins loading order is defined by the actual alphabetical order of the plugin subfolders and the plugins files, and finally by the order you define your decorated functions.

It would be useful to have something that tells Pyrogram in which order to exactly load plugins. Maybe something like setuptools' MANIFEST.in file.

PR #202, opened 2019-01-16
Smart plugins enhancements

delivrance:
From #183 and #197


delivrance:
Little mess (ignore): I accidentally committed changes on client.py with a different line separator.

PR #206, opened 2019-01-27
client.join_chat() now returns pyrogram.Chat instead of MTProto Update

bakatrouble:


pyrogram/client/methods/chats/join_chat.py
```diff
@@ -36,7 +37,7 @@ def join_chat(self,
         match = self.INVITE_LINK_RE.match(chat_id)
 
         if match:
-            return self.send(
+            chat = self.send(
```
delivrance: Here we could just use `Chat._parse_chat_chat` or `Chat._parse_channel_chat depending on the result...

pyrogram/client/methods/chats/join_chat.py
```diff
@@ -53,8 +54,10 @@ def join_chat(self,
                 access_hash=resolved_peer.chats[0].access_hash
             )
 
-            return self.send(
+            chat = self.send(
```
delivrance: ...and `Chat._parse_channel_chat` only here.

pyrogram/client/types/user_and_chats/chat.py
```diff
@@ -209,3 +209,12 @@ def _parse_full(client, chat_full: types.messages.ChatFull or types.UserFull) ->
                 parsed_chat.invite_link = full_chat.exported_invite.link
 
         return parsed_chat
+
+    @staticmethod
+    def _parse_mtproto_chat(client, chat):
+        if isinstance(chat, types.Chat):
+            return Chat._parse_chat_chat(client, chat)
+        elif isinstance(chat, types.User):
+            return Chat._parse_user_chat(client, chat)
```
delivrance: Since parsing User won't ever happen here (we are joining chats either with links or usernames) this method can be completely avoided (see previous code reviews).

bakatrouble: Why not leave this method just in case (for example) someone would want to parse MTProto chat types while working with raw updates?

delivrance: @bakatrouble Not sure. I don't quite like having something around that's not used, mainly because in the current state I'm already using each method singularly here and there. Anyway, these methods get messy pretty fast and get no docs whatsoever and could change anytime without notice, people should not rely on them. But if you want it we can keep it as `_parse_chat` with a comment to remember this is not used anywhere, but could be in case I go through the next project refactor iter.

PR #208, opened 2019-01-29
Add client.set_username() method

bakatrouble:


pyrogram/client/methods/users/set_username.py
```diff
@@ -0,0 +1,46 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Pyrogram.  If not, see <http://www.gnu.org/licenses/>.
+
+from pyrogram.api import functions
+from ...ext import BaseClient
+
+
+class SetUsername(BaseClient):
+    def set_username(self,
```
delivrance: Nice feature to have, but you forget that channel/supergroup usernames can also be changed.
We need two separate methods to distinguish the cases:

- `update_user_username(username: str)` (or simply update_username): for users only
- `update_chat_username(chat_id: int or str, username: str)`: for channels/supergroups only

Methods named `update_*` will also handle the removal of usernames by passing either an empty string or None explicitly; this will be explained in the docstrings. Alternatively we can have `set_*` and `remove_*` methods which do one thing only, set/update and remove. I personally prefer the first variant.

delivrance:
While `update_chat_username` feels nice as name, I can't say the same for `update_user_username` as "user" is repeated. I'm not sure whether to keep it this way or just have `update_username`, also because this will affect *only* the current user and adding "user" in the method name could (wrongly) hint you could change any user username.

PR #210, opened 2019-01-29
added missing await for send_audio

MBRCTV:


MBRCTV:
i have no clue why there is so many commits.. but i only changed one file..

delivrance:
It looks like you messed up with git in your own fork. Not a problem here, I'll just squash merge

MBRCTV:
Thanks Dan 👍

PR #216, opened 2019-02-10
Fixed loss of thumbnails after media editing.

23rd:
For some reason, message editing has completely missed thumbnails.


delivrance:
Thanks.

PR #221, opened 2019-02-21
Add bot_token argument

### LINKED ISSUE #123: Allow specifying custom file names for bot sessions
Currently, the session file name is named after the bot id, not really nice in case we wan't to tell multiple sessions apart. Having the possibility to specify a custom file name (just like user auths do) would be nice.

bakatrouble:
Closes #123.
Old behavior (bot token as session name) is still remaining with warning, but should be removed in future

### DIFF
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -293,6 +293,7 @@ class Client(Methods, BaseClient):
         try:
             if self.user_id is None:
                 if self.bot_token is None:
+                    self.is_bot = False
                     self.authorize_user()
                 else:
                     self.is_bot = True
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -29,6 +29,7 @@ import struct
 import tempfile
 import threading
 import time
+import warnings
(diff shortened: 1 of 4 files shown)


pyrogram/client/ext/base_client.py
```diff
@@ -68,7 +68,7 @@ class StopTransmission(StopIteration):
     }
 
     def __init__(self):
-        self.bot_token = None
+        self.is_bot = False
```
delivrance: Better set this to `None` because... (read next review)

pyrogram/client/client.py
```diff
@@ -284,11 +295,12 @@ def start(self):
                 if self.bot_token is None:
                     self.authorize_user()
```
delivrance: ...because you are relying on `self.is_bot = False` here. Better be explicit and set it to False here instead.

delivrance:
Beside the small review, I'd change nothing else. Nice job

bakatrouble:
> Beside the small review, I'd change nothing else. Nice job

Done.

PR #222, opened 2019-02-24
Added phone code hash to Client constructor

sybrpnk:


pyrogram/client/client.py
```diff
@@ -505,6 +507,7 @@ def authorize_bot(self):
     def authorize_user(self):
         phone_number_invalid_raises = self.phone_number is not None
         phone_code_invalid_raises = self.phone_code is not None
+        phone_code_hash_invalid_raises = self.phone_code_hash is not None
```
sybrpnk: Actually that's not really needed as I did not make a special exception

pyrogram/client/client.py
```diff
@@ -527,53 +530,56 @@ def default_phone_number_callback():
...
-                        self.api_hash
+            if phone_code is None or phone_code_hash is None:
+                try:
+                    r = self.send(
+                        functions.auth.SendCode(
+                            self.phone_number,
+                            self.api_id,
+                            self.api_hash
+                        )
                     )
```
sybrpnk: There are some huge changes because I have wrapped phone code hash getting block in the if statement

PR #227, opened 2019-03-09
Add ability to forward messages as copies

### LINKED ISSUE #172: Allow to forward a copy of the message (without the forward header)
Forwarded messages have headers on top of them telling from whom were originated. Telegram X and Plus Messenger have an option to forward a copy without the forward header.

bakatrouble:
Closes #172

### DIFF
--- a/pyrogram/client/methods/messages/forward_messages.py
+++ b/pyrogram/client/methods/messages/forward_messages.py
@@ -27,8 +27,10 @@ class ForwardMessages(BaseClient):
     def forward_messages(self,
                          chat_id: Union[int, str],
                          from_chat_id: Union[int, str],
-                         message_ids: Iterable[int],
-                         disable_notification: bool = None) -> "pyrogram.Messages":
+                         message_ids: Union[int, Iterable[int]],
+                         as_copy: bool = False,
+                         disable_notification: bool = None,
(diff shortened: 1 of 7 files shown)


pyrogram/client/methods/messages/forward_messages.py
```diff
@@ -59,35 +69,53 @@ def forward_messages(self,
...
-                from_peer=self.resolve_peer(from_chat_id),
-                id=message_ids,
-                silent=disable_notification or None,
-                random_id=[self.rnd_id() for _ in message_ids]
+        if as_copy:
+            sent_messages = []
+            for chunk in [message_ids[i:i+200] for i in range(0, len(message_ids), 200)]:
+                messages = self.get_messages(chat_id=from_chat_id, message_ids=chunk)  # type: pyrogram.Messages
+                for message in messages.messages:
+                    sent_messages.append(message.forward(
```
delivrance: Flood waits will break forwards in case of too many messages sent as-copies in a short period. We should handle such errors here.

delivrance:
I've added `send_cached_media` in the develop branch (details at https://github.com/pyrogram/pyrogram/commit/b3ef53a29a8dc153fa6f5f5576f79526897d428e). While useful for anyone, we could use that here too for all the applicable media messages as a first step to make this code cleaner.

PR #229, opened 2019-03-14
Hide Forward Message Headers

Rsplwe:


delivrance:
Thanks, but no, sorry. I'm not going to merge anything related to this before official apps update their TL schema. This is not public yet, telegram has not made any announcement and all of this can change anytime until official apps are done with the development of the feature.

For now you can just use the `experiments` branch, which already adds those two constructors to the L95 schema and works fine.

PR #230, opened 2019-03-16
Make use of __slots__ for efficiency

delivrance:


delivrance:
A brief test revealed that RAM usage dropped by around 42%

PR #231, opened 2019-03-16
Show TOS on signup just once while program is running

bakatrouble:


delivrance:
Looks good enough for now, thanks! I was thinking about an extra argument to make clients completely silent. We'll keep this for later, maybe once we refactor the whole config loader since is a mess to work with right now.

PR #233, opened 2019-03-20
Added convenience methods

xsolinsx:
message.reply_animation
message.reply_audio
message.reply_cached_media
message.reply_chat_action
message.reply_contact
message.reply_document
message.reply_game
message.reply_inline_bot_result
message.reply_location
message.reply_media_group
message.reply_photo
message.reply_poll
message.reply_sticker
message.reply_venue
message.reply_video
message.reply_video_note
message.reply_voice
message.edit_caption
message.edit_media
message.edit_reply_markup
message.pin

fixed send_document docstrings while doing so
uniformed function declaration of send_poll

### DIFF
--- a/pyrogram/client/methods/messages/send_document.py
+++ b/pyrogram/client/methods/messages/send_document.py
@@ -60,7 +60,7 @@ class SendDocument(BaseClient):
                 pass an HTTP URL as a string for Telegram to get a file from the Internet, or
                 pass a file path as string to upload a new file that exists on your local machine.
 
-            thumb (``str``):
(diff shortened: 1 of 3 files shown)


delivrance:
To keep the example snippets as short as possible I'm removing `reply_to_message_id=message.message_id` and `quote=True` to all reply_* methods except `reply` itself.

This will result in cleaner examples like `message.reply_audio(audio)`, as opposed to `message.reply_audio(audio, quote=True)`, thus giving an extra felling of the convenience these methods bring.

PR #237, opened 2019-04-14
Allow setting default command prefix through Filters.command_prefix

mendelmaleh:
Most people will anyway use the same prefix(es) for all commands, might as well set it once globally. Also, this way plugins will work with your preferred prefix, without extra configuration.


mendelmaleh:
Apparently [partials](https://docs.python.org/3.7/library/functools.html#functools.partial) exist, thanks @delivrance :+1:

delivrance:
Correct, as I explained already in https://t.me/c/1355792138/685. I don't think it is worth polluting the filters this way and I was about to give you an example. For your case you can make use of partials from `functools` this way:

```python
from functools import partial
from pyrogram import Client, Filters

app = Client("pyrogrambot")

dot_command = partial(Filters.command, prefix=".")

@app.on_message(dot_command("start"))
def _(c, m):
    print(m)

app.run()
```

PR #240, opened 2019-04-20
getpass() for 2FA password

codenoid:


delivrance:
@codenoid It actually used to be like that, but some people reported it was breaking things with their terminals and I got rid of it. Also, I don't think getpass can be made async, can it? Because we also want to use it in the asyncio branch for consistency.

PR #241, opened 2019-04-24
Use str or bytes for callback_data and CallbackQuery.data

JosXa:


pyrogram/client/types/bots/callback_query.py
```diff
@@ -79,7 +79,7 @@ def __init__(
         self.chat_instance = chat_instance
         self.message = message
         self.inline_message_id = inline_message_id
-        self.data = data
+        self.data: str = str(data, "utf-8")
```
delivrance: Neat! TIL.

PR #248, opened 2019-06-01
Add Filter for Callback_Query.data

ColinTheShark:
Make Filtering for specific callback_query's easier.


delivrance:
Now it's good enough for me, as we discussed: https://t.me/PyrogramLounge/68251

PR #268, opened 2019-07-02
Add message.web_page attributes

mendelmaleh:


pyrogram/client/types/messages_and_media/webpage.py
```diff
@@ -0,0 +1,192 @@
...
+    """A webpage preview
+
+    Parameters:
+        id (``str``):
+            Unique identifier for this webpage.
+
+        url (``str``):
+            Full URL for this webpage.
+
+        display_url (``str``):
```
delivrance: Required and optional arguments are like in the raw type, good.

pyrogram/client/types/messages_and_media/webpage.py
```diff
@@ -0,0 +1,192 @@
...
+            Display URL for this webpage.
+
+        type (``str``, *optional*):
+            Type of webpage preview, known types (at the time of writing) are:
+            `article`, `photo`, `gif`, `video` and `document`,
+            `telegram_user`, `telegram_bot`, `telegram_channel`, `telegram_megagroup`.
```
delivrance: While I strongly prefer having comprehensive docs, in this case we have no way of telling what all the possible values are (unless maybe we look at official clients/libs). This approach is good enough, except the style you used. Maybe it's better to use \*italic* instead, like I did in other places.

mendelmaleh: f7bfd55

mendelmaleh: Is this ok?

PR #277, opened 2019-07-11
Create a new update_profile method to update a users own profile

kalmengr:


JosXa:
There's a conflicting PR in https://github.com/pyrogram/pyrogram/pull/336. Could you guys discuss and collaborate to combine your versions?

delivrance:
Thanks @JosXa for pointing out. Not sure what's better, one method to change all profile details or three different methods for each. Probably the former.

Issue #163, opened 2018-11-21
There's not enough control over the authorization process


delivrance:
Currently, [Client.start()](https://docs.pyrogram.ml/pyrogram/Client#pyrogram.Client.start) takes care of every single step in the authorization process. The method is also quite big itself and by modularizing it not only makes it more maintainable but also allows the single login steps to be used independently, thus enabling more complex use cases that wouldn't be possible with start() alone.

delivrance:
This issue should gain some priority.

PR #281, opened 2019-07-14
Start by steps

adolfaka:
I propose to divide the start() and authorize_user() methods into separate steps for use in web applications. For example, in Flask https://github.com/adolfaka/pyroflask


delivrance:
Thanks! I'll audit the PR more thoroughly asap and eventually adapt it better. I'm seeing many things that need to be changed, like some methods names (update, download) and the fact you call load_settings() on each method. There should be a cleaner way to achieve the same result.

delivrance:
This PR is going to address #163 (I'm mentioning the issue just in case someone stumbles upon it so it can see the PR opened as well).

adolfaka:
Thank you for paying attention to this problem.  It will be great if you implement it in your project.

PR #285, opened 2019-07-20
Add support for downloading files to file pointers

### LINKED ISSUE #284: Support for in-memory file downloads
## Checklist
- [ ] I believe the idea is awesome and would benefit the library.
- [ ] I have searched in the issue tracker for similar requests, including closed ones.
- [ ] Currently you can only download to file by path, not to BytesIO/anyotherIO or just bytes
- [ ] Feature parity with plain bot API libraries

## Description
An argument `out` for message.download, which is annotated by io.IOBase and downloads to specified file-like object. If both `file_name` and `out` are specified, throw exception

OctoNezd:
Fixes #284

### DIFF
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -15,7 +15,7 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with Pyrogram.  If not, see <http://www.gnu.org/licenses/>.
-
+import io
 import logging
 import math
 import mimetypes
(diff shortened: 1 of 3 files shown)


alemigliardi:
This is quite old now, is this never going to happen? Is the implementation unsatisfactory? This is pretty relevant and makes handling files cleaner. If issue are merge conflicts I can fix those, I'd really like this added

PR #290, opened 2019-07-28
Don't print None

mendelmaleh:


mendelmaleh:
not needed anymore d875298

PR #292, opened 2019-07-29
Improvements for message.command

mendelmaleh:
Enables easier and standard parsing, for quote wrapped args, etc.
Filters.command now has a posix argument, and the separator argument was
removed. shlex.split() works similar to having before separator=None.


delivrance:
As we discussed in the group, I've updated the filter. We still have the issue about unmatched quotes, though.

PR #296, opened 2019-08-06
Added support for bot_token inside config.ini file

trenoduro:
If 'bot_token' is explicitly defined, than that is used, otherwise if there is a bot token in the config.ini file that is used.
bot_token is set to None if not present


delivrance:
Thanks for your contribution.

pyrogram/client/client.py
```diff
@@ -1228,6 +1229,11 @@ def send(self, data: TLObject, retries: int = Session.MAX_RETRIES, timeout: floa
     def load_config(self):
         parser = ConfigParser()
         parser.read(str(self.config_file))
+        
+        if self.bot_token:
+            pass
```
YoilyL: Why not `if not self.bot_token`? am I missing something?

PR #297, opened 2019-08-06
Add better support for nested entities (both for HTML and Markdown)

Midblyte:
As per the title...


delivrance:
Great, thank you. I'm going to fix some tiny issues

PR #305, opened 2019-08-19
Add bound method for get_common_chats()

ColinTheShark:
Meanwhile I've taken the liberty and removed whitespaces from empty lines.


delivrance:
Well I assume that was vscode. Pycharm could add those back when I reformat in future

PR #310, opened 2019-08-28
support None as config_file to avoid trying to read any file

lesinigo:
There is currently no way to avoid trying to read a `config_file` because in the `Client` constructor the relevant parameter gets immediately passed through `pathlib.Path()`and then on to `ConfigParser().read()`.

This PR adds support for `Client(config_file=None)` to avoid trying to read anything.

### DIFF
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -157,7 +157,7 @@ class Client(Methods, BaseClient):
             where Pyrogram will store your session files. Defaults to the parent directory of the main script.
 
         config_file (``str``, *optional*):
-            Path of the configuration file. Defaults to ./config.ini
+            Path of the configuration file, or None to avoid trying to read one. Defaults to ./config.ini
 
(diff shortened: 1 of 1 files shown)


delivrance:
@lesinigo Hi. I can't see why this is useful since files that don't exist are silently ignored and the code strictly checks for sections existence before trying to read from the `parser`.

For instance, using `Client(...)` alone would fail because no api key was provided, whereas `Client(..., api_id=..., api_hash=...)` works fine already.

lesinigo:
@delivrance it's the opposite of non existent files: the PR makes possible to avoid reading an existent file; this behavior must be explicitly requested and if not used it does not change the standard behavior.

delivrance:
@lesinigo I fail to understand the reasoning behind your proposal. If the config file exists, why don't you want it to be read? Also remember that arguments passed to `Client` already take precedence over what is read from the config file.

PR #322, opened 2019-10-04
Add mark_chat_unread() method

ColinTheShark:
Title says all. Mark a chat as unread.

https://t.me/PyrogramChat/123932


delivrance:
@ColinTheShark Can you confirm this works without the need of an `InputDialogPeer` object?
If that works, it means the server accepts either types (as it happens in a bunch of other places) and thus I think we can accept the method having `chat` instead of `dialog` in its name for convenience.

delivrance:
Also, what about marking a chat as read instead? The raw method allows to do that too, but the issue here is the misleading name.

ColinTheShark:
> Can you confirm this works without the need of an `InputDialogPeer` object?

I haven't explicitly tested against an `InputDialogPeer` object but I can confirm that [my snippet](https://github.com/ColinTheShark/Pyrogram-Snippets/blob/master/Snippets/unread.py) <sup>(same raw method)</sup> works both in private- and group chats.

---

> Please also add a convenience method to `user_and_chats/chat.py`

You mean a bound method, right? If so, I added that already :)

---

> Also, what about marking a chat as read instead?

While `read_history` can mark up to a specific point (or the latest message if no max_id is given), `MarkDialogUnread` just marks as unread. I have tried with `unread=False` as an argument, which didn't remove the unread status. I'm sure there is a raw method to mark a chat as read without using `read_history`, but I believe Clients just use the raw equivalent to `read_history`.

pyrogram/client/types/user_and_chats/chat.py
```diff
@@ -754,3 +754,23 @@ def export_invite_link(self):
         """
 
         return self._client.export_invite_link(self.id)
+
+    def mark_chat_unread(self):
```
delivrance, line 758: When we add a new method we usually name it like `app.action_on_type`, when it's a bound method we remove "type" because it's redundant (bad: `chat.mark_chat_unread`, good: `chat.mark_unread`).

delivrance (approved):

delivrance (approved):

PR #323, opened 2019-10-04
Add basic constants

adolfaka:
It will be great to add basic constants with restrictions by Telegram


delivrance:
Sorry, rejecting this for now. Mainly because some of the constants actually change over time and because some can be retrieved anytime with `help.getConfig`. Also, you did put a low effort in adding them as it looks like a copy-paste from a telegram bot framework, hence many of them are invalid for Pyrogram (upload/download size is 1500MB, not 50/20, webhooks don't exist in Pyrogram, to name a few).

PR #336, opened 2019-10-29
Add update_bio() method

ColinTheShark:
>>> app.update_bio("New Bio!")
    True

### DIFF
--- a/pyrogram/client/methods/users/update_bio.py
+++ b/pyrogram/client/methods/users/update_bio.py
@@ -34,7 +34,7 @@ class UpdateBio(BaseClient):
 
         Parameters:
             bio (``str`` | ``None``):
-                Bio to set. "" (empty string) or None to remove it.
+                Bio to set. Pass "" (empty string) or None to remove it.
 
         Returns:
             ``bool``: True on success.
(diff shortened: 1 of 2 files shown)


pyrogram/client/methods/users/update_bio.py
```diff
@@ -0,0 +1,54 @@
...
+        self,
+        bio: Union[str, None]
+    ) -> bool:
+        """Update your own bio.
+
+        This method only works for users, not bots. Bot bio must be changed via BotFather.
+
+        Parameters:
+            bio (``str`` | ``None``):
+                Bio to set. "" (empty string) or None to remove it.
```
JosXa: How about

Bio to set. Use an empty string or `None` to remove it.

? :)

JosXa:
There's a conflicting PR in https://github.com/pyrogram/pyrogram/pull/277. Could you guys discuss and collaborate to combine your versions?
In particular, "bio" seems to be the more slangy term for the "about text" of a user profile.

delivrance:
Merged #277, closing this.
I think it's better to have all details editable with a single method as in the raw API.

PR #339, opened 2019-11-06
Poll ID should be INT

trenoduro:
Otherwise stop_poll() raises AtributeError thile trying to convert a string to bytes


delivrance:
The poll id is a string to make it similar to the bot API, we should edit `stop_poll` and convert the string id into integer instead with `id=int(poll.id)` and keep the rest untouched.

PR #346, opened 2019-11-27
filter messages coming from the channel connected to the chat

xsolinsx:
maybe filter's name is not the best but that's the only thing I could think of


delivrance:
`linked_channel` is better, since that's how official clients call it as well!

PR #343, opened 2019-11-15
add missing file_ref in set_chat_photo

Sunda001:
fix empty file reference

PR #347, opened 2019-11-27
add file_ref to convenience methods where it was missing

xsolinsx:


JosXa:
This and https://github.com/pyrogram/pyrogram/pull/343/files make me think that there might be other places where the argument is missing 🤔

xsolinsx:
@JosXa Very likely

PR #367, opened 2020-02-04
Fix RPCError 400 QUIZ_CORRECT_ANSWER_EMPTY

trenoduro:
If correct_option_id == 0 then None is sent instead of 0

### DIFF
--- a/pyrogram/client/methods/messages/send_poll.py
+++ b/pyrogram/client/methods/messages/send_poll.py
@@ -110,7 +110,7 @@ class SendPoll(BaseClient):
                         public_voters=not is_anonymous or None,
                         quiz=type == "quiz" or None
                     ),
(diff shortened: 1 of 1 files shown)


pyrogram/client/methods/messages/send_poll.py
```diff
@@ -110,7 +110,7 @@ def send_poll(
                         public_voters=not is_anonymous or None,
                         quiz=type == "quiz" or None
                     ),
-                    correct_answers=[bytes([correct_option_id])] if correct_option_id else None
+                    correct_answers=[bytes([correct_option_id])] if not correct_option_id == None else None
```
alissonlauffer: I think would be better if you use this:
```python
correct_answers=None if correct_option_id is None else [bytes([correct_option_id])]
```
Both codes should work as expected, but the above one is more consistent (The same style is used on other parts of the library).

PR #373, opened 2020-02-26
Add message.get_poll_option_votes()

bakatrouble:
Requires testing and design review

### DIFF
--- a/pyrogram/client/types/messages_and_media/message.py
+++ b/pyrogram/client/types/messages_and_media/message.py
@@ -21,11 +21,14 @@ from typing import List, Match, Union
 
 import pyrogram
 from pyrogram.api import types
+from pyrogram.api.functions.messages import GetPollVotes
 from pyrogram.client.types.input_media import InputMedia
 from pyrogram.errors import MessageIdsEmpty
 from .contact import Contact
 from .location import Location
 from .message_entity import MessageEntity
+from .poll_option import PollOption
+from .poll_option_vote import PollOptionVote
 from ..messages_and_media.photo import Photo
 from ..object import Object
 from ..update import Update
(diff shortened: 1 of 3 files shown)


pyrogram/client/types/messages_and_media/message.py
```diff
@@ -3108,3 +3111,45 @@ def pin(self, disable_notification: bool = None) -> "Message":
             message_id=self.message_id,
             disable_notification=disable_notification
         )
+
+    def get_poll_option_votes(self, option: PollOption):
+        """Get votes for attached public poll option
```
JosXa, line 3116: Maybe "Get votes for an option attached to a public poll"? A bit easier to comprehend

JosXa:
Is there a corresponding batch method for this?
I'd like to know if users are gonna have to make an outbound request for every single option if they want to retrieve all vote counts. 

**If there is a batch call:**
--> Please implement it

**If not:**
--> Think about whether it would be possible to abstract this in a nice way. Thinking about asyncio, firing a bunch of requests in parallel is possible using `await asyncio.gather(*tasks)`. But maybe it could also be abstracted as a list of requests for `client.send(...)` ("true batching").

pyrogram/client/types/messages_and_media/message.py
```diff
@@ -3108,3 +3111,45 @@ def pin(self, disable_notification: bool = None) -> "Message":
...
+        Parameters:
+            option (``pyrogram.PollOption``):
+                The poll option you want to get votes list for.
+
+        Returns:
+            List of :obj:``PollOptionVote``
+
+        Raises:
+            RPCError: In case of a Telegram RPC error.
+            ``ValueError``: If the message doesn't contain a poll or attached poll is anonymous
```
JosXa, line 3132: Inconsistent docstring style:

> RPCError

but

> \``ValueError\``

pyrogram/client/types/messages_and_media/poll_option_vote.py
```diff
@@ -0,0 +1,44 @@
...
+
+        date (``int``):
+            Date the vote was submitted in Unix time.
+    """
+
+    def __init__(
+        self,
+        *,
+        client: "pyrogram.BaseClient" = None,
+        user: 'pyrogram.User',
```
JosXa, line 38: Is it normal to use "type.Annotations" style here? Couldn't there be a relative import from the underlying module to use the concrete type instead of using `pyrogram.X` for that?

JosXa:
A convenience method on `Poll` would be cool: https://github.com/pyrogram/pyrogram/blob/master/pyrogram/types/messages_and_media/poll.py
Right?

PR #397, opened 2020-05-04
Update MTProto vs BotAPI - Download/upload section

CyanBook:
Based on the recent changes on Telegram API (https://core.telegram.org/bots/api#getfile), bots can now download and upload files up to 20MB. Then Pyrogram docs have to change.


zeroone2numeral2:
[As far as I can see](https://core.telegram.org/bots/api#inputfile), bot API's `InputFile` still allow to upload files up to 50 mb of size (unless they are pictures). The method you link is used to pull files, and returns a [`File`](https://core.telegram.org/bots/api#file) object - which is 20 mb of size at max (it's coherent with what the documentation says about downloading files)

CyanBook:
Nevermind, my error 😅

I received false information from a friend. sorry!

PR #400, opened 2020-05-06
Relative path error

alessiocelentano:
When a script is run from a directory other than its parent, it cannot find the config.ini. With these small changes it seems to work


delivrance:
You are adding some inconsistencies here by not taking care of `workdir` as well (one line above `config_file`).

PR #404, opened 2020-05-11
set_web_page_preview method and a bugfix

SuperCz1:
- Fixed this bug: https://t.me/pyrogramchat/169553
- Added the client.set_web_page_preview() method, to set the default status of "disable_web_page_preview" parameters, as client.set_parse_mode()

### DIFF
--- a/pyrogram/client/methods/messages/send_message.py
+++ b/pyrogram/client/methods/messages/send_message.py
@@ -115,6 +115,9 @@ class SendMessage(BaseClient):
                             [InlineKeyboardButton("Docs", url="https://docs.pyrogram.org")]
                         ]))
         """
+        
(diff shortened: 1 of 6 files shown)


delivrance:
Thanks, but you should have split this into two separate PRs.

pyrogram/client/methods/messages/send_message.py
```diff
@@ -115,6 +115,9 @@ def send_message(
                             [InlineKeyboardButton("Docs", url="https://docs.pyrogram.org")]
                         ]))
         """
+        
+        if disable_web_page_preview is None:
```
delivrance, line 119: You are missing a few other places where `disable_web_page_preview` is used. Please search for all occurrences and add this.

PR #422, opened 2020-06-15
Send a warning instead throwing an exeption when message type is not …

fgallaire:
…compatible with as_copy
As messages copies are often done in batches, send a warning and continue the work is a far better behavior (so much user friendly).

### DIFF
--- a/pyrogram/client/types/messages_and_media/message.py
+++ b/pyrogram/client/types/messages_and_media/message.py
@@ -2687,9 +2687,9 @@ class Message(Object, Update):
         """
         if as_copy:
             if self.service:
-                log.warning("Unable to copy service messages, message_id: {} from chat.id: {}".format(self.message_id, self.chat.id))
+                log.warning(f"Service messages cannot be copied. chat_id: {self.chat.id}, message_id: {self.message_id}")
             elif self.game and not self._client.is_bot:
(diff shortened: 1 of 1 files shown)


fgallaire:
@delivrance do you prefer a `no_fail` boolean for retro-compatibility reason ?

delivrance:
I tried to forward three messages, two texts and one service in the middle; the server successfully forwards the two texts. I approve this.

delivrance:
> @delivrance do you prefer a `no_fail` boolean for retro-compatibility reason ?

I'd rather keep the code simpler and mention in the release note about this possible breaking change. Thanks

pyrogram/client/types/messages_and_media/message.py
```diff
@@ -2684,12 +2687,10 @@ def forward(
         """
         if as_copy:
             if self.service:
-                raise ValueError("Unable to copy service messages")
-
-            if self.game and not self._client.is_bot:
-                raise ValueError("Users cannot send messages with Game media type")
-
-            if self.text:
+                log.warning("Unable to copy service messages, message_id: {}".format(self.message_id))
```
delivrance, line 2690: Maybe also mention the source chat id (`self.chat.id`) and not only the message id.

PR #427, opened 2020-06-20
Don't exclude pinned messages in get_dialogs call

vtronko:
Currently, when you issue a `get_dialogs` call with default parameters (precisely, 
`pinned_only=False`) — pinned messages are implicitly excluded. This PR
amends that.

### DIFF
--- a/pyrogram/client/methods/chats/get_dialogs.py
+++ b/pyrogram/client/methods/chats/get_dialogs.py
@@ -75,8 +75,7 @@ class GetDialogs(BaseClient):
                     offset_id=0,
                     offset_peer=types.InputPeerEmpty(),
                     limit=limit,
-                    hash=0,
-                    exclude_pinned=True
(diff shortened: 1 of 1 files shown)


delivrance:
It's often hard to make api design decisions because people will always find something that doesn't fit well for their purposes.

Here, your change is limiting in case we want to only fetch dialogs that are not pinned. The raw API actually provides two distinct methods, `messages.GetPinnedDialogs` and `messages.GetDialogs`. For convenience, pyrogram merged both methods into one. Maybe we should keep two separate convenience method too: `get_pinned_dialogs`, which would return a list of pinned dialogs only and `get_dialogs` (which should be turned into a generator and have `iter_dialogs` deprecated) with an optional parameter `exclude_pinned`. What do you think?

JosXa:
I like the `exclude_*` approach.

PR #436, opened 2020-07-03
Hide password on prompt

slntopp:
Hiding password sudo-like using default Python module: getpass
Works on all OS(Windows, Linux, Mac)

### DIFF
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -24,6 +24,7 @@ import shutil
 import tempfile
 import threading
 import time
+from getpass import getpass
 from configparser import ConfigParser
 from hashlib import sha256, md5
 from importlib import import_module, reload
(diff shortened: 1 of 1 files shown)


delivrance:
@slntopp Thanks for the contribution, however, this was already proposed and refused due to 1) `getpass` being problematic on some terminal locales and 2) there is no `async` variant of `getpass`. If we can find a solution to at least the second issue I can reconsider this.

slntopp:
Ah so, then sorry, I haven't seen the previous issue about.
Could you link an issue and write here what is the problem between async and get pass, where do you use inputs asyncly?

delivrance:
The problem is that `getpass` has no async interface, but neither does `input`. Using `getpass` as it is would block the event loop. I managed to turn `input` async using executors and maybe the same can be achieved for `getpass`.
https://github.com/pyrogram/pyrogram/blob/dd9b55f256c306bb26f7985e0cab07dd2be278a0/pyrogram/client/ext/utils.py#L86-L92.

PR #437, opened 2020-07-07
support for both sync and async filters

HasibulKabir:


pyrogram/client/ext/dispatcher.py
```diff
@@ -188,8 +188,12 @@ async def update_worker(self, lock):
 
                             if isinstance(handler, handler_type):
                                 try:
-                                    if handler.check(parsed_update):
-                                        args = (parsed_update,)
+                                    if callable(handler.filters) and asyncio.iscoroutinefunction(handler.filters.__call__):
```
delivrance, line 191: `callable(handler.filters)` is not actually needed. Filter objects in the current API always have `__call__` defined, if they don't they are not proper filters and pyrogram will raise an exception when accessing `handler.filters.__call__`, which is better than having it silently ignore updates.

delivrance: I think we should remove all of these checks here and have `handler.check` actually check for coroutine functions. It also helps removing one repeated line `args = (parsed_update,)`.

HasibulKabir: sure.

pyrogram/client/handlers/deleted_messages_handler.py
```diff
@@ -46,5 +46,5 @@ class DeletedMessagesHandler(Handler):
     def __init__(self, callback: callable, filters=None):
         super().__init__(callback, filters)
 
-    def check(self, messages):
-        return super().check(messages[0])
+    async def check(self, messages):
```
delivrance, line 49: You checked for other usages. Good!

pyrogram/client/handlers/handler.py
```diff
@@ -16,14 +16,22 @@
...
+
+        if callable(self.filters):
+            if asyncio.iscoroutinefunction(self.filters.__call__):
+                return (await self.filters(update))
+
+            else:
+                return (self.filters(update))
+
+        else:
+            return (True)
```
delivrance, line 37: Why all these parentheses?

HasibulKabir: I have checked after this commit. I have removed last line and added  parentheses in return.

PR #447, opened 2020-07-30
set_my_commands method

CrisMystik:
Added the client.set_my_commands() method as the SetMyCommands botapi function.
* Async version: https://github.com/CrisMystik/pyrogram/tree/pr1-asyncio

### DIFF
--- a/pyrogram/client/methods/users/set_my_commands.py
+++ b/pyrogram/client/methods/users/set_my_commands.py
@@ -46,7 +46,7 @@ class SetMyCommands(BaseClient):
                 functions.bots.SetBotCommands(
                     commands=[
                         types.BotCommand(command=command, description=description)
-                        for command, description in commands
+                        for command, description in commands.items()
                     ]
                 )
             )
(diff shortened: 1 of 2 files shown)


delivrance:
Maybe we should mimic the bot api arguments and types since this is a bot-only feature, but instead of `set/get_my_commands` we should call it `set/get_bot_commands` because it's weird for user identities.

delivrance:
Also, there's no need to duplicate the work for asyncio since currently is mostly a matter of adding async/await keywords where needed (I do this myself when merging from the develop branch).

pyrogram/client/methods/users/set_my_commands.py
```diff
@@ -0,0 +1,53 @@
...
+from pyrogram.api import functions, types
+from ...ext import BaseClient
+
+
+class SetMyCommands(BaseClient):
+    def set_my_commands(
+        self,
+        commands: Dict[str, str]
+    ) -> bool:
+        """Update your bot commands list ("/" icon in official Telegram clients).
```
JosXa, line 30: > ("/" icon in official Telegram clients)

I'd honestly just leave it out.
If you wanna keep it, maybe shorten to `(slash-commands)` ?

JosXa (approved):
Nice!

PR #456, opened 2020-08-07
Add support from pathlib objects

CyanBook:
Add support to pathlib object, by transforming them to a path string when detected.

Also, for some unknown reason an old commit already pushed is inside in this pull request. sry


CyanBook:
Changing every file is bad practise, since i can modify save_file directly

PR #457, opened 2020-08-07
Update save_file function to support pathlib objects

CyanBook:
save_file can now support pathlib objects.

As open() also supports pathlib object already, combining the same if statement for string paths and pathlib objects works perfectly fine

### DIFF
--- a/pyrogram/client/client.py
+++ b/pyrogram/client/client.py
@@ -28,7 +28,7 @@ import time
 from configparser import ConfigParser
 from hashlib import sha256, md5
 from importlib import import_module
-from pathlib import Path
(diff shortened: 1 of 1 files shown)


pyrogram/client/client.py
```diff
@@ -1788,7 +1788,7 @@ def save_file(
 
         part_size = 512 * 1024
 
-        if isinstance(path, str):
+        if isinstance(path, str) or isinstance(path, PurePath):
```
delivrance, line 1791: Use `isinstance(path, (str, PurePath))` instead. The function allows multiple types to be checked with a single call.

PR #467, opened 2020-08-20
Update chosen_inline_result.py

MarcosRandulfeGarrido:
Fix some syntax errors that are causing errors on python pip package distribution


MarcosRandulfeGarrido:
I think that is a small but important and urgent fix, I dowload pyrogram through pip and I couldn't do a simple hello world with these problem.

PR #472, opened 2020-08-24
[filters] added entity filters

Accipiter7:


Accipiter7:
@delivrance  closing because it would be better to implement entity filter that takes entity type as parameters

PR #473, opened 2020-08-24
one time filter check instead of checking everytime

HasibulKabir:


delivrance:
Unfortunately, this requires adding extra code for class-based filters (see filters.user, filters.chat) in order to explicitly tell the Filter whether its \_\_call\_\_ method is async or not, which is something I don't quite like since I might want to document how class-based filters work (that is, I don't want people to deal with an attribute that is totally useless to them as it is just an internal detail).

delivrance:
To clarify, class-based filters don't use filters.create, hence `"iscoroutinefunction": inspect.iscoroutinefunction(func)` will not exist for them and you need to manually add it each time.

delivrance:
About the new changes: filters != callback, they are different things and what I said above still holds.

PR #477, opened 2020-08-27
update_chat_notifications new high-level method

CyanBook:
update_chat_notifications is now a high-level method

- it has bound methods for chat object
- it supports unix time, datetime and timedelta

### DIFF
--- a/pyrogram/methods/chats/update_chat_notifications.py
+++ b/pyrogram/methods/chats/update_chat_notifications.py
@@ -1,5 +1,4 @@
 from typing import Union
-from datetime import datetime, timedelta
 
 from pyrogram.raw.functions.account import UpdateNotifySettings
 from pyrogram.raw.types import InputNotifyPeer, InputPeerNotifySettings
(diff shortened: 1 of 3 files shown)


pyrogram/methods/chats/update_chat_notifications.py
```diff
@@ -0,0 +1,75 @@
...
+from pyrogram.scaffold import Scaffold
+
+
+class UpdateChatNotifications(Scaffold):
+    async def update_chat_notifications(
+        self,
+        chat_id: Union[int, str],
+        show_previews: bool = None,
+        silent: bool = None,
+        mute_until: Union[int, datetime, timedelta] = None
```
delivrance, line 15: This adds inconsistencies with the rest of the API, which currently only allows unix timestamps to be passed as argument.

delivrance (changes requested):

pyrogram/methods/chats/update_chat_notifications.py
CyanBook: Now it accepts only unix time

delivrance (approved):
Ok as an initial support. I'm going to round some edges and maybe add a new convenience type as we probably need it.

mrksbgg:
god bless pyro devs 🙏

arynyklas:
no way

rojvv:
Don't trash the comments so bad that Dan won't touch Pyrogram for another two years.

ALiwoto:
wow amazing

KurimuzonAkuma:
Time to finish with my fork?

asnct:
yay

PR #500, opened 2020-10-01
changed emogis colors to the correct values and added a offset to sta…

gouravkmar2:
…rt time


pyrogram/session/auth.py
```diff
@@ -119,7 +121,9 @@ async def create(self):
                     server_nonce=server_nonce,
                     new_nonce=new_nonce,
                 ).write()
-
+                if s>start*10:
```
delivrance, line 124: What is this?

delivrance:
Hi. This PR is weird, I'm not sure why you were fiddling with the auth internals. Plus, a PR should contain changes about one single self contained context instead of multiple different areas. Closing.

PR #502, opened 2020-10-02
Fix Typo

KILR007:
"redisribution" to "redistribution"


delivrance:
Thanks, however the typo is not mine since the document comes from https://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types.
You might want to search for its repository and propose the change there as well!

PR #503, opened 2020-10-02
Documentation improvement

jaivanti:
📄Added a content improvement line in the start to make document attractive!
peace✌


delivrance:
Thanks for your suggestion, however I have worded the readme carefully and I don't think this change fits well.

PR #510, opened 2020-10-04
Code Improvements & Errors

pokurt:
Dan said to pr to Main (former master) there fore this is where I will open my pull requests.
- Fixed format string in `compiler/api/compiler.py`
- Swapped if/else to remove 'pass' (empty if body)

Changes in `pyrogram/dispatcher.py`:
- removed 'i' variable and replaced with _ cause my ide was screaming and it wont effect the code itself (pls laugh i made a joke)
- moved 'parser' variable closer to the usage

Merged nested 'if' conditions in `pyrogram/sync`

Changes in `pyrogram/utils`:
- Merged nested 'if' conditions
- changed 'parsed_messages' list loop into list comprehension

like i said my ide screams so  replaced variable with _ in `pyrogram/connection/connection.py`

Changes in `pyrogram/crypto/aes.py`:
- hoisted for/while loops out of nested contionals
- replaced range(int, var) to range(var) why is an int there anyway?

### DIFF
--- a/pyrogram/methods/advanced/save_file.py
+++ b/pyrogram/methods/advanced/save_file.py
@@ -154,7 +154,7 @@ class SaveFile(Scaffold):
                     chunk = fp.read(part_size)
 
                     if not chunk:
-                        if not is_big:
+                        if not is_big and not is_missing_part:
                             md5_sum = "".join([hex(i)[2:].zfill(2) for i in md5_sum.digest()])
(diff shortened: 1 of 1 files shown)


delivrance:
@pokurt Hi. I think you synced your fork with pyrogram:master and all your changes are now lost. You should have used a different branch for your PR.

PR #511, opened 2020-10-06
Allow dialog methods to fetch archived chats

### LINKED ISSUE #362: Allow get/iter_dialogs to fetch archived chats

kandnub:
This PR closes #362 

get_dialogs_count/get_dialogs/iter_dialogs: Added "archived_only" kwarg.

types.Dialog: Added "is_archived" boolean attribute.

### DIFF
--- a/pyrogram/methods/chats/get_dialogs.py
+++ b/pyrogram/methods/chats/get_dialogs.py
@@ -32,7 +32,8 @@ class GetDialogs(Scaffold):
         self,
         offset_date: int = 0,
         limit: int = 100,
-        pinned_only: bool = False
+        pinned_only: bool = False,
+        archived_only: bool = False
     ) -> List["types.Dialog"]:
         """Get a chunk of the user's dialogs.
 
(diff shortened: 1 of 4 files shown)


delivrance:
Hi. Two issues:
- You are using a boolean value where an integer is required. `archived_only` is boolean but the method works with integer folder ids.
- This method is not that trivial to deal with. Telegram works with an optional folder id. If you don't pass it you'll get all dialogs, if you pass 0 you get the non-archived dialogs, if you pass 1 you get the archived dialogs. With your solution we are missing one case.

I believe the whole method needs a rewrite to allow an easier implementation and usage, in order to allow every possible case. Maybe pinned dialogs should also be fetched using a separate convenience method?

kandnub:
You're correct about those issues, I used `archived_only` as a boolean since it's a subclass of int and to pass None to it directly to fetch all dialogs (which is dirty), but that doesn't seem suitable for `folder_id` anymore, plus it would fail with `GetPinnedDialogs` since `folder_id` isn't optional there. My apologises for not testing all the possible scenarios.

I also looked a bit into `GetPinnedDialogs`, `GetDialogs` and `GetPeerDialogs` (new `get_dialog` method maybe), having separate methods for them should be better. Maybe move the common dialog parsing code into a function outside the class and separate the methods in their classes respectively rather making a new file to avoid boilerplate?

Also, currently `GetDialogs` has a limit of only retrieving 100 dialogs the most, perhaps we could call `iter_dialogs` inside `get_dialogs` to make better use of the `limit` argument and rename `pinned_only` to `exclude_pinned` to match the raw method more?

Wip-Sama:
so... this is not implemented?

PR #518, opened 2020-10-20
ADD: Small recomendations in some parts of files in /pyrogram about P…

NicoGutierrezGarrido:
…EP8 standar


ColinTheShark:
This feels like a not needed PR.
Dan likes to have the code 120 chars, and as long as this is consistent across the whole code-base, it can stay that way.

Context https://t.me/pyrogramlounge/233963

PR #519, opened 2020-10-20
Add in-memory uploads for send_media_group

odysseusmax:


fgallaire:
@delivrance we need to have this

PR #533, opened 2020-11-02
Add custom caption argument to forward messages

M-S-2-7:
add a custom_caption argument to forward_messages
it works if as_copy equals True


delivrance:
Hi. The whole `as_copy` feature has been superseded by the new `copy_message` method.

PR #539, opened 2020-11-17
add-inline-query-audio

JollyGoal:
Added audio support for inline query
credits: DogeSaveTheGod

### DIFF
--- a/pyrogram/types/inline_mode/inline_query_result_audio.py
+++ b/pyrogram/types/inline_mode/inline_query_result_audio.py
@@ -1,0 +1,132 @@
+#  Pyrogram - Telegram MTProto API Client Library for Python
+#  Copyright (C) 2017-2020 Dan <https://github.com/delivrance>
+#
+#  This file is part of Pyrogram.
+#
+#  Pyrogram is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published
+#  by the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  Pyrogram is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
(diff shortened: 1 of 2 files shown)


delivrance:
This is a Bot API specific feature, I believe we should mimic it to keep the same behaviour and interface.

pyrogram/types/inline_mode/inline_query_result_audio.py
```diff
@@ -0,0 +1,132 @@
...
+        input_message_content (:obj:`~pyrogram.types.InputMessageContent`):
+            Content of the message to be sent instead of the photo.
+    """
+
+    def __init__(
+            self,
+            audio_url: str,
+            title: str,
+            duration: int = 0,
+            voice: bool = False,
```
delivrance, line 72: The Bot API has a different type for voice results

pyrogram/types/inline_mode/inline_query_result_audio.py
```diff
@@ -0,0 +1,132 @@
...
+            performer: str = "",
+            mime_type: str = "audio/ogg",
```
delivrance, line 74: There's no mime_type parameter in the Bot API, is this really useful?

pyrogram/types/inline_mode/inline_query_result_audio.py
JollyGoal: I did not get that, why is the type for inline audio is different?

pyrogram/types/inline_mode/inline_query_result_audio.py
JollyGoal: I thought it would be useful in future, anyways it has default value so it won't cause errors or something.
I will delete mime_type param and commit ASAP

pyrogram/types/inline_mode/inline_query_result_audio.py
JollyGoal: What is confusing at this state? I'll change it

delivrance: The bot API has `InlineQueryResultAudio` as well as `InlineQueryResultVoice`. Two different classes.

delivrance (approved):

PR #550, opened 2020-11-25
New method - get_media_group() 

Legenda24:
Method to get all media group messages knowing only one. Special thank to Danipulok (https://github.com/Danipulok) 

"Problems":
1) Docs
Doubt how to write correctly so I marked those lines with  "#TODO" + examples. 

2) Should it raises error in case it's passed "wrong" (not related to media group) message_id or just return an empty list? 

UPD(5.12.2020):
I just noticed that other media groups have media_group_id attribute too (playlists etc.)
So now I'm 100% sure this method supports all types of media groups(playlists, albums etc.)  as it should be :)

### DIFF
--- a/pyrogram/methods/messages/get_media_group.py
+++ b/pyrogram/methods/messages/get_media_group.py
@@ -1,0 +1,38 @@
+import logging
+from typing import Union, List
+
+from pyrogram.scaffold import Scaffold
+from pyrogram.types import list
+
+log = logging.getLogger(__name__)
+
+
+class GetMediaGroup(Scaffold):
+    async def get_media_group(
+        self,
+        chat_id: Union[int, str],
+        message_id: int
+    ) -> List["types.Message"]:
+        """Get media group 
+
+        Parameters:
+            chat_id (``int`` | ``str``):
+                Unique identifier (int) or username (str) of the target chat.
+                For your personal cloud (Saved Messages) you can simply use "me" or "self".
+                For a contact that exists in your Telegram address book you can use his phone number (str).
+
+            message_id (int):
+                #TODO Pass one message of target media get_media_group
+        Returns:
+            List of :obj:`~pyrogram.types.Message`
+        Raises:
+            #TODO ValueError: In case message isn't related to media group. 
+        """
+        messages = await self.get_messages(chat_id, [msg_id for msg_id in range(message_id-9, message_id+10)], replies=0)
+
+        if messages[9].media_group_id:
+            media_group_id = messages[9].media_group_id
+        else:
+            raise ValueError("Message isn't related to media group") #TODO
+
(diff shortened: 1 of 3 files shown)


Danipulok:
Well, I can say that we were testing a few options and it was the fastest one and pretty the easiest one to read.

And I also think that the name should be _get_album()_, because if there's method that will get all the files in one file group (like a few audiofiles together or a file .txt or other files) that one should be called _get_media_group()_.

pyrogram/methods/messages/get_media_group.py
```diff
@@ -0,0 +1,56 @@
...
+            #TODO ValueError: In case message isn't related to media group. 
+        """
+        messages = await self.get_messages(chat_id, [msg_id for msg_id in range(message_id-9, message_id+10)], replies=0)
+
+        if messages[9].media_group_id:
+            media_group_id = messages[9].media_group_id
+        else:
+            raise ValueError("Message isn't related to media group") #TODO
+
+        return list.List(msg for msg in messages if msg.media_group_id == media_group_id)
```
delivrance, line 56: You can import `List` directly from `pyrogram.types`

pyrogram/methods/messages/get_media_group.py
```diff
@@ -0,0 +1,56 @@
...
+            List of :obj:`~pyrogram.types.Message`
+        Raises:
+            #TODO ValueError: In case message isn't related to media group. 
+        """
+        messages = await self.get_messages(chat_id, [msg_id for msg_id in range(message_id-9, message_id+10)], replies=0)
+
+        if messages[9].media_group_id:
+            media_group_id = messages[9].media_group_id
+        else:
+            raise ValueError("Message isn't related to media group") #TODO
```
delivrance, line 54:
``` python
media_group_id = messages[9].media_group_id

if media_group_id is None:
    raise ...
```

looks better.

pyrogram/methods/messages/get_media_group.py
Legenda24: Yes, but it will be mismatching between typing List and pyrogram List

delivrance: Not really, pyrogram's List is an actual list, it exists only for pretty print purposes. The only issue I see here is that you have to rename either one so it's ok to keep it like so.

Legenda24: I am better off with the present state of things.

delivrance (changes requested):
All is good. Could you add the friendly bound method `Message.get_media_group` as well?
chat and message id will be taken from the message instance itself, so this method will have no parameter.

Legenda24:
Don't sure, so more like no. Tbh this is my very first experience "contributing" smth in open-source project ;)
I'd better leave this job for more "knowing" people

PR #565, opened 2020-12-19
Added url Filter

logicfool:
Url filter

It would be easier for people working with urls to use this filter and check if that message has a url or not so they wont have to go all the way through entities and check the type and then use if else

right now I just used entities to go through all of them but will read the whole pyrogram code and modify this later on to make it better then it is right now

### DIFF
--- a/pyrogram/filters.py
+++ b/pyrogram/filters.py
@@ -414,6 +414,17 @@ web_page = create(web_page_filter)
 
 # endregion
 
+#region url_filter
+async def url_filter(_,__, m: Message):
+    return bool(True if True in ([True if (entities.type == 'url') else False for entities in m.entities] if m.entities else [False]) else False)
+
+
+has_url = create(url_filter)
+"""Filter messages that contains Url"""
+
+
+#endregion
+
 # region poll_filter
 async def poll_filter(_, __, m: Message):
     return m.poll


ChenZehong13 (approved):

zeroone2numeral2:
Hmm, why not just adding a filter that filters messages based on an user-specified entity? Such as python-telegram-bot's [`Filters.entity()`](https://python-telegram-bot.readthedocs.io/en/stable/telegram.ext.filters.html#telegram.ext.filters.Filters.entity), to which one could pass a list of [`pyrogram.types.MessageEntity`](https://docs.pyrogram.org/api/types/MessageEntity#pyrogram.types.MessageEntity). Stopping at checking whether a message has urls or not feels unnecessarily limited

logicfool:
Yeah that can be done as well and that would be better but as everything else like common entities are present in filters directly and url is used by alot of programmers so just thought why not make a common filter for urls.

The method you said looks clean will try to impliment that till then this can be considered as a quick fix for all those who want to skip those nested if else's for checking if a url is present in their text or not.

zeroone2numeral2:
Uhm, I think quick fixes be part of the user's code instead of being embedded directly in the library

logicfool:
Yeah that's true will close this Pull Request.

ChenZehong13 (commented):
？

ColinTheShark:
Hi @logicfool, Your implementation of a URL filter looks quite messy in my opinion. The other filters we have in Pyrogram aren't quite for MessageEntities, but for attributes already easily accessible. These are convenient for the user as they don't have to do long if-chains anymore. A URL filter for the MessageEntity isn't really beneficial, as a User still has to check the individual Entities to check whether or not the URL is correct or expected. It would be a lot easier to use a [basic URL RegEx](https://regex101.com/r/ARxpnQ/1) <sup>(I made this in about a minute)</sup>.

@ChenZehong13 From your commit history in your own repositories and your stars I take you don't work with Python or Pyrogram often or at all. Why do you leave Reviews and comments that contribute literally nothing to the PR at hand? <sup>That's a rhetorical question</sup>

logicfool:
Ah yes I understand better not put it in the main library..

Issue #570, opened 2020-12-24
Can't download_media on Photos


alemigliardi:
## Checklist
- [x] I am sure the error is coming from Pyrogram's code and not elsewhere.
- [x] I have searched in the issue tracker for similar bug reports, including closed ones.
- [x] I ran `pip3 install -U https://github.com/pyrogram/pyrogram/archive/master.zip` and reproduced the issue using the latest development version.

## Description
`client.download_media` fails on Photos if no explicit file name is given.
This happens because `Photos` types lack both a `mime_type` field and a `file_name` field, making it impossible to guess the file extension. Pyrogram 1.0.7 correctly guessed it, I assume because `Photos` can only be `.jpg` anyway?

## Steps to Reproduce
Send a photo, try to download it with `await client.download_media(message)`

## Traceback
```
Traceback (most recent call last):
  File "/home/<me>/venv/pyrogram/lib/python3.8/site-packages/pyrogram/dispatcher.py", line 208, in handler_worker
    await handler.callback(self.client, *args)
  File "/home/<me>/userbot/selfbot/plugins/media.py", line 68, in download_attached_media
    fname = await client.download_media(message)
  File "/home/<me>/venv/pyrogram/lib/python3.8/site-packages/pyrogram/methods/messages/download_media.py", line 137, in download_media
    guessed_extension = self.guess_extension(mime_type)
  File "/home/<me>/venv/pyrogram/lib/python3.8/site-packages/pyrogram/client.py", line 1052, in guess_extension
    return self.mimetypes.guess_extension(mime_type)
  File "/usr/lib/python3.8/mimetypes.py", line 192, in guess_extension
    extensions = self.guess_all_extensions(type, strict)
  File "/usr/lib/python3.8/mimetypes.py", line 171, in guess_all_extensions
    type = type.lower()
AttributeError: 'NoneType' object has no attribute 'lower'
```

emanuelmendez:
I have the exactly same problem on 1.1.3

pokurt:
Same here. on some media. it works on others lol.

alemigliardi:
Fixed in 1.1.5

PR #571, opened 2020-12-24
Don't guess extension of Photos

alemigliardi:
Media of type `Photo` lacks `mime_type` : calling `guess_extension` on it results in an exception because its `mime_type` is `None`. Don't try to guess file extension of photos and just use `.jpg` (it didn't use the guessed extension anyway).

This should solve #570


alemigliardi:
Fixed in 1.1.5, Dan once again fixes fast and better than me

PR #585, opened 2021-01-09
Fix invalid inline message id

Ripeey:
Fix for inline message id invalid when target dc is not dc4 or current dc


delivrance:
Was that it? Well nice, from what people reported to me it seemed like it was just DC4's fault. Thank you.

PR #587, opened 2021-01-10
temp fix for annoying warning

Krishna-Singhal:
Since setup via pyrogram, "No plugins loaded from *" is irritating sometimes. isnt it ?
that's why tried to fix this issue, 🙁

### DIFF
--- a/pyrogram/client.py
+++ b/pyrogram/client.py
@@ -706,8 +706,6 @@ class Client(Methods, Scaffold):
             include = plugins["include"]
             exclude = plugins["exclude"]
 
-            count = 0
-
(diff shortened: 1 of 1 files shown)


ColinShark:
What is the goal of this PR exactly? Would be nice to have some additional information on what this does instead of just "temp fix".

Krishna-Singhal:
> What is the goal of this PR exactly? Would be nice to have some additional information on what this does instead of just "temp fix".

[Oops, I Forgot](https://github.com/pyrogram/pyrogram/pull/587#issue-552258303)

PR #592, opened 2021-01-16
Add method client.copy_media_group

Danipulok:
A new method client.copy_media_group. 
You pass a chat_id and a single ``message_id``, and ``media_group`` that contains message with this ``message_id`` will be copied.

It was useful for me so I thought I could PR this.
I wanted to create it as flexible as possible, so attribute ``captions`` became a little mess inside, but IMO it's gonna be easy in use in exchange.

Big thanks for Legenda24 for helping me [(github](https://github.com/Legenda24), [Telegram ](https://t.me/legendax24))

### DIFF
--- a/pyrogram/methods/messages/copy_media_group.py
+++ b/pyrogram/methods/messages/copy_media_group.py
@@ -1,0 +1,134 @@
+#  Pyrogram - Telegram MTProto API Client Library for Python
+#  Copyright (C) 2017-2021 Dan <https://github.com/delivrance>
+#
+#  This file is part of Pyrogram.
+#
+#  Pyrogram is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published
+#  by the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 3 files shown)


pyrogram/methods/messages/copy_media_group.py
```diff
@@ -0,0 +1,134 @@
...
+            else:
+                raise TypeError("Message with this type can't be copied.")
+
+            media = utils.get_input_media_from_file_id(file_id=file_id)
+            multi_media.append(
+                raw.types.InputSingleMedia(
+                    media=media,
+                    random_id=self.rnd_id(),
+                    **await self.parser.parse(
+                        captions[i] if type(captions) is list and i < len(captions) and captions[i] is not None else
```
code-rgb, line 105: `type(captions) is list` can replaced with  `isinstance(captions, list)`

Danipulok: Well seems like a nice idea

delivrance (approved):

PR #597, opened 2021-01-17
on_error() decorator added

jonatan1609:
a new on_error decorator has been added.
you can use it via

```
@app.on_error()
def error_handler(client, exception_object):
    pass

```
or

```
@app.on_error(check_all=True)
def error_handler(client, exception_object):
    pass
```

same as 

`app.add_handler(handlers.error_handler.ErrorHandler(app, error_handler, True))`

In order to catch all errors raised on the application, not only RPCErrors.

### DIFF
--- a/pyrogram/handlers/__init__.py
+++ b/pyrogram/handlers/__init__.py
@@ -25,3 +25,4 @@ from .message_handler import MessageHandler
 from .poll_handler import PollHandler
 from .raw_update_handler import RawUpdateHandler
 from .user_status_handler import UserStatusHandler
+from .error_handler import ErrorHandler
--- a/pyrogram/handlers/error_handler.py
+++ b/pyrogram/handlers/error_handler.py
@@ -18,6 +18,7 @@
 
 import sys
 import asyncio
+import functools
 
 from pyrogram import errors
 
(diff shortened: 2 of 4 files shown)


pokurt:
this is one of the useless pr ive seen so far but you are damn creative on this

Sunda001:
> this is one of the useless pr ive seen so far but you are damn creative on this

not really useless

zeroone2numeral2:
@pokurt mind motivating why you consider this useless? I honestly see why configuring a callback that should run when one of your handlers raises an exception can be considered a useful feature to have. Maybe it might be considered out of the scope of a library such as pyrogram

@jonatan1609 a cool thing to have would be to tell the dispatcher whether to continue to propagate the update to the following handlers, if the exception is handled by an error handler

jonatan1609:
> @jonatan1609 a cool thing to have would be to tell the dispatcher whether to continue to propagate the update to the following handlers, if the exception is handled by an error handler 


Yup that's cool. I am waiting for @delivrance 's answer what does he think about it, if it will be okay I'll keep developing it and i will add docs as well.

jonatan1609:
Do not merge it yet, i need to fix some bugs.

pokurt:
lmao

jonatan1609:
What’s so fooni

PR #603, opened 2021-01-23
Fix to work with edit_message_caption works with reply_markup objects

CyanBook:
Since CallbackQuery.edit_message_text takes 4 arguments and CallbackQuery.edit_message_caption only 3, the reply_markup ends up to be the disable_web_page_preview one. Resolve this by specifying the argument name


code-rgb:
edit message caption reply_markup still missing for me

delivrance (approved):

PR #610, opened 2021-01-28
create a __main__ file to get pyrogram version easily

M-S-2-7:
you can get the pyrogram version easily by running this command:
`python -m pyrogram`

### DIFF
--- a/pyrogram/__main__.py
+++ b/pyrogram/__main__.py
@@ -1,0 +1,2 @@
+from . import version
+print('Pyrogram', version)


M-S-2-7:
oh sorry, the `version` variable should be `__version__`

jonatan1609 (changes requested):
change `version` to `__version__` by add a new commit :)

ColinShark:
I feel like this is a rather useless addition. Pyrogram itself is not a module to be called, but a library to be imported. Using the `-m` argument tells Python to start a module with the name passed.
If you need the exact version of Pyrogram, you can simply ask Pip or `import pyrogram`

```
> pip show pyrogram
Name: Pyrogram
Version: 1.1.13
[..]
```

```python
>>> import pyrogram
>>> pyrogram.__version__
'1.1.13'
```

Note: I'm not closing this, as I'd like to hear @delivrance's input on this, too.

delivrance:
> I feel like this is a rather useless addition. Pyrogram itself is not a module to be called, but a library to be imported. Using the `-m` argument tells Python to start a module with the name passed.
> If you need the exact version of Pyrogram, you can simply ask Pip or `import pyrogram`

I agree. Printing just the version doesn't justify the module usage given the other ways. What I expect from executing packages as modules is they to become interactive in some way, something like a small embedded application that lets you explore the library (not something I actually think to add).

PR #615, opened 2021-02-03
bot mention in groups

jonatan1609:
i have added option to mention a bot in a group by /command@username and not only /command,
it only works in a group since this is only relevant there.


jonatan1609 (commented):
Seems like it could be merged to the next version :D

jonatan1609:
Awesome 😃

PR #618, opened 2021-02-08
ignore null buttons in reply_markup.write()

naftali100:
this change is fixing a problem with building keyboard when the button is null
its happens when trying to copy message with unsupported button (login button) which is null 
https://t.me/pyrogramchat/291847 - my question about it


delivrance:
The issue should be fixed as LoginUrl buttons have been added.

PR #619, opened 2021-02-08
Fixed method `Client.get_media_group`

Danipulok:
Fixed a case in which get_media_group() raises an error when it shouldn't..

**Steps to reproduce:**
1. Create a new supergroup
2. Send any media group (for example, 3 photos). This way first message of `media_group` will have an `message_id` of 1.
3. Pass to the `Client.get_media_group()` `message_id` of the any message in this media group message and you'll not get media group, but an ValueError instead.

**Reason of the problem** 
So the `get_messages()` in[ get_media_group()](https://github.com/pyrogram/pyrogram/blob/master/pyrogram/methods/messages/get_media_group.py#L52) result `message_ids` will be:
`1                           - service message of creating a supergroup`
`2, 3, 4                    - media group messages`
`5, 6, 7, 8, 9, 11, 12 - empty messages`

And getting the target `media_group_id` is just getting 9th elem in the list of messages. But in this case it will neither get the right message nor media_group_id (because the 9th elem will not be the message related to media group).

**Solution**
So instead of just getting 9th item in messages list, we just check its len, which will either 
• return 19 messages if passed message_id is more than 9 (so everything is alright in old method) 
• or will have less then 19 messages (that would raise a problem in an old method), what's solved by the following `else` condition:
`media_group_id = messages[9].media_group_id if len(messages) == 19 else messages[message_id-1].media_group_id`, where message_id will be gotten correctly.

Also, thanks Legenda24 ([Telegram](t.me/legendax24), [GitHub](https://github.com/legenda24)) for forcing me fix this without extra request and helping with writing this explanation.


pyrogram/methods/messages/get_media_group.py
```diff
@@ -48,10 +48,15 @@ async def get_media_group(
         Raises:
             ValueError: In case the passed message id doesn't belong to a media group.
         """
-        # There can be maximum 10 items in a media group. 
-        messages = await self.get_messages(chat_id, [msg_id for msg_id in range(message_id - 9, message_id + 10)], replies=0)
+        messages = await self.get_messages(
+            chat_id=chat_id,
+            message_ids=[msg_id for msg_id in range(message_id - 9, message_id + 10)],
```
delivrance, line 53: This still breaks if you pass a low `message_id` (e.g. 2), as it will try to obtain messages with ids <= 0.

pyrogram/methods/messages/get_media_group.py
```diff
@@ -48,10 +48,15 @@ async def get_media_group(
...
+            replies=0
+        )
 
-        media_group_id = messages[9].media_group_id
+        # There can be maximum 10 items in a media group.
+        # The if/else condition to fix the problem of getting correct `media_group_id` when it has `message_id` less then 10.
+        media_group_id = messages[9].media_group_id if len(messages) == 19 else messages[message_id-1].media_group_id
```
delivrance, line 59: Maybe we should search again for the message having the same id we passed to the method instead of relying on indexes.

delivrance (changes requested):

pyrogram/methods/messages/get_media_group.py
Danipulok: When I tested this about 1 month ago, it worked perfectly.

The thing is, if you pass (for example):
_app.get_messages(chat_id=chat_id, [-5, 5])_
it returns a list only with positive message_ids.

So in this case it'd return List of Messages, where message_ids are [from '1' to '5'].
Telegram just doesn't return messages if you pass negative message_id.

pyrogram/methods/messages/get_media_group.py
Danipulok: That's true, but when we first created this method with Legenda24, he was very concious about the speed.

And by adding those conditions we still have the same speed and 100% method. (I really doubt Client API will ever return `Messages` with negative `message_id`s).

Legenda24: Anyway this case is very very rare. Never seen that somebody faces with it. (in main chat)

Danipulok: Well, but I personally faced this. 

And including the fact that's it's 'our' method, I want it to be as good as possible :D

Legenda24: at the moment it's good as possible

Danipulok: Well, if you mean this with this PR it will become as good as possible, then I totally agree.

Legenda24: > I want it to be as good as possible :D
>> its good as possible

Danipulok: >  if you mean that with this PR it will become as good as possible, then I totally agree.

Just like I said.

pyrogram/methods/messages/get_media_group.py
delivrance: Ok, what actually breaks is get_messages with a single negative id, which is not applied here.

pyrogram/methods/messages/get_media_group.py
Legenda24: I said it like 2-3 replies earlier. don't need to explicit what I meant

pyrogram/methods/messages/get_media_group.py
Danipulok: So then we should check if the input `message_id` is not <= 0, right?

Danipulok: Dan, please, review the commit.

delivrance (approved):

PR #630, opened 2021-03-04
Invite Links v2

delivrance:


delivrance:
I think we should rename all `Invite(s)` occurrences to `InviteLink(s)` for consistency with other older parts (export_chat_invite_link, Chat.invite_link).

delivrance:
While at it, we should also move the ability to check invite links out of `get_chat` and have, say, `check_chat_invite_link` instead.

pyrogram/methods/invites/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with Pyrogram.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from .delete_exported_chat_invite_link import DeleteExportedChatInviteLink
+from .delete_revoked_exported_chat_invite_links import DeleteRevokedExportedChatInviteLinks
+from .edit_exported_chat_invite_link import EditExportedChatInviteLink
+from .export_chat_invite_link import ExportChatInviteLink
+from .get_admins_with_invite_links import GetAdminsWithInviteLinks
```
ColinShark, line 24: `get_admins_with_invite_links.py` is still missing as I don't know what exactly is returned here. Trying to install and run this PR will raise and error, that python could not import this file.

delivrance:
Names are getting quite long here. We should try to shorten them:

Does "delete_exported" revoke them?
Does "delete_revoked_exported" delete revoked ones only?

Some proposals:

- ~delete_exported_chat_invite_link~ -> revoke_chat_invite_link
- ~delete_revoked_exported_chat_invite_link~ -> delete_chat_invite_link
- ~edit_exported_chat_invite_link~ -> edit_chat_invite_link
- ~export_chat_invite_link~ is "export" any meaningful now? maybe we should use "generate" since there can be more than one, so -> generate_chat_invite_link
- get_admins_with_invite_links looks ok, but where's the method to get all links from one admin?
- get_chat_invite_link_importers. Not sure how understandable is using import/export in this context. Importers are people that join via links, exporters are admins who generate links.
- get_exported_chat_invite_links, same here.

ColinShark:
> Names are getting quite long here. We should try to shorten them:

Yeah, it's a hassle to work with tbh xD

---
 
> Does "delete_exported" revoke them?

No, non-revoked links will raise an error `INVITE_REVOKED_MISSING`. Invite Links need to first be edited with `revoked=True`.

---

> Does "delete_revoked_exported" delete revoked ones only?

Yes, this will deleted all links of the specified admin that are already revoked. A quick way to clear the list of revoked links (as they will remain displayed in the UI).

---

> Some proposals: (I edited your bullet points to a numbered list)
> 
> 1. ~delete_exported_chat_invite_link~ -> revoke_chat_invite_link
> 2. ~delete_revoked_exported_chat_invite_link~ -> delete_chat_invite_link
> 3. ~edit_exported_chat_invite_link~ -> edit_chat_invite_link
> 4. ~export_chat_invite_link~ is "export" any meaningful now? maybe we should use "generate" since there can be more than one, so -> generate_chat_invite_link
> 5. get_admins_with_invite_links looks ok, but where's the method to get all links from one admin?
> 6. get_chat_invite_link_importers. Not sure how understandable is using import/export in this context. Importers are people that join via links, exporters are admins who generate links.
> 7. get_exported_chat_invite_links, same here.

1. Since `delete_exported_chat_invite_link` does not actually revoke links, but fails when trying to delete non-revoked links, calling this method `revoke_chat_invite_link` would be counter-intuitive.
2. Instead of `delete_revoked_exported_chat_invite_link` or `delete_chat_invite_link`, what about `clear_revoked_invite_links`?
3. `edit_chat_invite_link` sounds good. This method allows to edit revoked links with `revoke=False` to reinstate them, too.
4. `generate_chat_invite_link` also sounds like a good idea.
5. "but where's the method to get all links from one admin?" That would be `get_exported_chat_invite_links`. We could probably ditch the `exported` and make it just `get_chat_invite_links`?
6. I have no idea how understandable "importers" is in the context of invite links and I don't know what a good substitution could be. `get_chat_invite_link_users`? `get_joiners`?
7. See 5.

delivrance:
Apparently, you can push to PR fork branches only if it's the fork owner who opens the PR in the first place and allows maintainer edits.  Since it was me who opened the PR there seems to be no way to toggle "allow edits from maintainer".
I'm merging this and will apply my changes separately.

ColinShark:
GitHub is weird. At least I got all this to work out.

Keep in mind that the current version of this PR will fail, because `invites/get_admins_with_invite_links` is missing. Please make sure to include it as soon as possible.

delivrance:
Closed due to PR updating issues (reopened in #639).

PR #644, opened 2021-03-19
A new attribute

jonatan1609:
The attribute  that used to determine whether the message contains media or not, is now named , and instead of it - the attribute  currently contains the media itself - no matter what type of media it is. [photo, location, contact, etc]


zeroone2numeral2:
This looks a bit too redundant, what are some use cases where it is preferable to have a native "`has_something`" boolean value to check instead of checking whether the `something` attribute is actually populated? As far as I know, truthiness checks on non-boolean values don't cause that much overhead either (correct me if I'm wrong)

jonatan1609:
> This looks a bit too redundant, what are some scenarios where it is preferable to have a native "`has_something`" boolean value to check instead of checking whether the `something` attribute is actually populated, if it's going to exist regardless? As far as I know, truthiness checks on non-boolean values don't cause that much overhead either (correct me if I'm wrong)

You might be right, maybe if i delete has_media ( and then has_media will be the same as bool(media), if media:, etc. ) it wl be better..

jonatan1609:
But anyway regardless to the feature (that Message.media should be the media itself, not a bool), the attribute should be named has_media and not media, at least as of now when media contains `True/False`

zeroone2numeral2:
Oh, my bad - I incorrectly assumed that `message.media` was containing the media object itself, but it's actually a boolean value. I personally would simply make `message.media` contain the media object itself, this way people can still use `if message.media` to check whether the message is a media or not, and use that attribute to do whatever they want with the media itself

jonatan1609:
> Oh, my bad - I incorrectly assumed that `message.media` was containing the media object itself, but it's actually a boolean value. I personally would simply make `message.media` contain the media object itself, this way people can still use `if message.media` to check whether the message is a media or not, and use that attribute to do whatever they want with the media itself - message.is_media would still be a bit redundant

yeah, *exactly* what I said here:

> > This looks a bit too redundant, what are some scenarios where it is preferable to have a native "`has_something`" boolean value to check instead of checking whether the `something` attribute is actually populated, if it's going to exist regardless? As far as I know, truthiness checks on non-boolean values don't cause that much overhead either (correct me if I'm wrong)
> 
> You might be right, maybe if i delete has_media ( and then has_media will be the same as bool(media), if media:, etc. ) it will be better..
> 
> I can just let the end-user determine if there's a media or not by himself, simply by
> 
> ```python
> if message.media:
>     print(message.media)
> ```
> 
> For case that user sent a document to the bot -- a document object will be printed.

zeroone2numeral2:
If figured that out from that comment indeed

I've added to my comment this bit btw:

> Anyway, I suspect `message.media` being a bool instead of the media object is a design choice - maybe because media objects are way too different to actually make `message.media` useful to use

jonatan1609:
> I figured that out from that comment indeed
> 
> I've added to my comment this bit btw:
> 
> > Anyway, I suspect `message.media` being a bool instead of the media object is a design choice - maybe because media objects are way too different to actually make `message.media` useful to use

I don't think, ok, so well if it's a design choice then name the attribute `has_media` to avoid confusions.

penn5 (changes requested):
Useless and waste of time

jonatan1609:
> Useless and waste of time

Just like you -.-

jonatan1609:
i removed has_media since 

```python
if message.media:
    pass
```

has the same effect.

PR #651, opened 2021-03-26
fix code quality issues

withshubh:
### Description
Hi :wave: I work at [DeepSource](https://deepsource.io), I ran DeepSource analysis on the forked copy of this repo and found some interesting [code quality issues](https://deepsource.io/gh/withshubh/pyrogram/issues/?category=recommended) in the codebase, opening this PR so you can assess if our platform is right and helpful for you.

### Summary of changes
- Remove unnecessary use of comprehension
- Refactor useless `else` block in the loop 
- Revert "Refactor useless `else` block in the loop"
- Remove methods with unnecessary super delegation.
- Refactor unnecessary `else` / `elif` when `if` block has a `continue` statement
- Access attribute directly
- Remove unnecessary f-string
- Remove assignment of variable to itself
- Remove redundant `None` default
- Add .deepsource.toml to continuously analyze the repo for code quality issues

### DIFF
--- a/pyrogram/session/session.py
+++ b/pyrogram/session/session.py
@@ -358,7 +358,7 @@ class Session:
 
         # Call log.debug twice because calling it once by appending "data" to the previous string (i.e. f"Kind: {data}")
         # will cause "data" to be evaluated as string every time instead of only when debug is actually enabled.
-        log.debug("Sent:")
+        log.debug("Sent:\n%s", message)
         log.debug(message)
 
         payload = await self.loop.run_in_executor(
--- a/pyrogram/session/session.py
+++ b/pyrogram/session/session.py
@@ -244,8 +244,7 @@ class Session:
             if msg.seq_no % 2 != 0:
                 if msg.msg_id in self.pending_acks:
                     continue
-                else:
-                    self.pending_acks.add(msg.msg_id)
+                self.pending_acks.add(msg.msg_id)
 
             if isinstance(msg.body, (raw.types.MsgDetailedInfo, raw.types.MsgNewDetailedInfo)):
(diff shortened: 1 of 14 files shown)


withshubh:
Hi @delivrance Please have a look at this PR

HitaloSama:
Wow, DeepSource really is pretty good.

pyrogram/session/session.py
```diff
@@ -359,7 +358,7 @@ async def _send(self, data: TLObject, wait_response: bool = True, timeout: float
 
         # Call log.debug twice because calling it once by appending "data" to the previous string (i.e. f"Kind: {data}")
         # will cause "data" to be evaluated as string every time instead of only when debug is actually enabled.
-        log.debug(f"Sent:")
+        log.debug("Sent:")
```
alissonlauffer, line 361: Seeing back these lines remember me a code change that was made in https://github.com/pyrogram/pyrogram/commit/76985faa118835e1b8ee2e5fb8ce51631d90ffd4, that changes `log.debug(f"Sent:\n{message}")` to a twice log.debug call, to avoid transforming the `Message` object into a string unnecessary. But the correct fix for that issue should be:
```python
log.debug("Sent:\n%s", message)
```

instead of calling log.debug twice.

withshubh: made the suggested change!

alissonlauffer: https://github.com/pyrogram/pyrogram/blob/a48d27f501b73eee842384723c33db298ce1e6c3/pyrogram/session/session.py#L237-L238

Can you apply it there too? And there are some log.* using f-strings which could be refactored with logging string formatter (I've found them with `grep -rEI 'log\.[a-z_]+?\(f"'`).

PR #653, opened 2021-03-27
Create delete_account.py

TAOTAO5:
add delete_account()

### DIFF
--- a/pyrogram/methods/users/delete_account.py
+++ b/pyrogram/methods/users/delete_account.py
@@ -1,0 +1,50 @@
+#  Pyrogram - Telegram MTProto API Client Library for Python
+#  Copyright (C) 2017-2021 Dan <https://github.com/delivrance>
+#
+#  This file is part of Pyrogram.
+#
+#  Pyrogram is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published
+#  by the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  Pyrogram is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


jonatan1609 (changes requested):
In the function signature you can just do
`reason: str = ""`
So you could do
`reason=reason` instead of `reason=reason or ""`

and you also need to inherit DeleteAccount in https://github.com/pyrogram/pyrogram/blob/4ebdc39fdd885c6a9dd7289e474d28cbe972f609/pyrogram/methods/users/__init__.py

pyrogram/methods/users/delete_account.py
```diff
@@ -0,0 +1,50 @@
...
+#  along with Pyrogram.  If not, see <http://www.gnu.org/licenses/>.
+
+from pyrogram import raw
+from pyrogram.scaffold import Scaffold
+
+class DeleteAccount(Scaffold):
+    def delete_account(
+        self,
+        reason: str = None
+    ) -> bool:
```
AndrielFR, line 26: Use `reason: str = ""`.

pyrogram/methods/users/delete_account.py
```diff
@@ -0,0 +1,50 @@
...
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with Pyrogram.  If not, see <http://www.gnu.org/licenses/>.
+
+from pyrogram import raw
+from pyrogram.scaffold import Scaffold
+
+class DeleteAccount(Scaffold):
+    def delete_account(
+        self,
+        reason: str = None
```
AndrielFR, line 25: Use `reason: str = ""`.

TAOTAO5: ok👌, changes have been made

jonatan1609:
👌

PR #657, opened 2021-03-31
new method set_bot_commands

jonatan1609:
```python
app.set_bot_commands(
    [
         pyrogram.types.BotCommand("start", "start the bot"),
         pyrogram.types.BotCommand("stop", "stop the bot")
    ]
)
```
in order to remove all commands, we can pass `[]` or `None`.

### DIFF
--- a/compiler/docs/compiler.py
+++ b/compiler/docs/compiler.py
@@ -392,6 +392,7 @@ def pyrogram_api():
             CallbackQuery
             GameHighScore
             CallbackGame
+            BotCommand
         """,
         input_media="""
         Input Media


pyrogram/types/bots_and_keyboards/bot_commands_list.py
```diff
@@ -0,0 +1,39 @@
...
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with Pyrogram.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import List
+
+from pyrogram import types
+from ..object import Object
+
+
+class BotCommandsList(Object):
```
delivrance, line 25: I don't think we need this. We can simply pass/return `List[BotCommand]`.

delivrance (changes requested):

pyrogram/types/bots_and_keyboards/bot_commands_list.py
jonatan1609: I think I've done that.

jonatan1609:
I fixed a SyntaxError, thanks to Github Actions 😆

pyrogram/methods/bots/set_bot_commands.py
```diff
@@ -0,0 +1,83 @@
...
+from pyrogram import types
+from pyrogram.scaffold import Scaffold
+
+
+class SetBotCommands(Scaffold):
+    async def set_bot_commands(
+            self,
+            commands: Union[
+                List[types.BotCommand],
+                List[List[str]],
```
delivrance, line 31: Here we can enforce the use of `List[types.BotCommand]` only to make it less confusing and also remove unnecessary checks in the code below.

delivrance (changes requested):

pyrogram/methods/bots/set_bot_commands.py
jonatan1609: Done

delivrance:
I updated some parts since we don't need all that complexity.

PR #673, opened 2021-04-30
Small improvements for handlers and filters

FeeeeK:
These changes allow passing kwargs to handlers from filters.
Here is an example of how this can be used:
```python
import re
from pyrogram import filters, Client

client = Client(...)

def regex_filter(pattern):
    def check(flt, client, message):
        match = re.match(flt.pattern, message.text)
        if match:
            return {"text": match.group(1)}

    return filters.create(
        check,
        "RegexFilter",
        pattern=pattern,
    )


@client.on_message(regex_filter(r"/say (.*)"))
def handler(client, message, text):
    message.reply(text)


client.run()
```
image

### DIFF
--- a/pyrogram/dispatcher.py
+++ b/pyrogram/dispatcher.py
@@ -20,6 +20,7 @@ import asyncio
 import inspect
 import logging
 from collections import OrderedDict
+from functools import partial
 
 import pyrogram
 from pyrogram import utils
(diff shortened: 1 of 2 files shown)


Mahesh0253:
```python
from pyrogram import filters, Client

client = Client(...)

@client.on_message(filters.regex(r"/say (.*)"))
def handler(client, message):
    text = message.matches[0].group(1)
    message.reply(text)

client.run()
```

You can use this method

PR #681, opened 2021-05-13
Add method to set the History TTL

ColinShark:
`app.set_history_ttl(chat_id, period)` to automatically delete messages after either a day (`86400`) or a week (`604800`). `0` to disable.

### DIFF
--- a/pyrogram/methods/chats/set_history_ttl.py
+++ b/pyrogram/methods/chats/set_history_ttl.py
@@ -1,0 +1,83 @@
+#  Pyrogram - Telegram MTProto API Client Library for Python
+#  Copyright (C) 2017-2021 Dan <https://github.com/delivrance>
+#
+#  This file is part of Pyrogram.
+#
+#  Pyrogram is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published
+#  by the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  Pyrogram is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with Pyrogram.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Union, Optional
+
+from pyrogram import raw
+from pyrogram import types
+from pyrogram.scaffold import Scaffold
+
+
+class SetHistoryTTL(Scaffold):
+    async def set_history_ttl(
+        self, chat_id: Union[int, str], period: int
+    ) -> "types.Message":
+        """Set the time-to-live for the chat history.
+
+        Parameters:
+            chat_id (``int`` | ``str``):
+                Unique identifier (int) or username (str) of the target chat.
+
+            period (``int``):
+                The time-to-live for the chat history.
+                Either 86000 for 1 day or 608000 for 1 week.
+                0 (zero) can be used to disable it.
+                Other values are not supported by Telegram.
+
+        Returns:
+            :obj:`~pyrogram.types.Message`: On success, the generated Service Message is returned.
+
+        Example:
+            .. code-block:: python
+
+                # One Day
+                app.set_history_ttl("PyrogramChat", 86400)
+
+                # A Week
+                app.set_history_ttl("PyrogramChat", 604800)
+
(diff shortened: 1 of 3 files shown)


pyrogram/methods/chats/set_history_ttl.py
```diff
@@ -0,0 +1,83 @@
...
+
+        Raises:
+            ValueError: In case the chat_id doesn't belong to a supergroup or the time-to-live isn't 1 day or week.
+        """
+
+        peer = await self.resolve_peer(chat_id)
+        if not isinstance(peer, raw.types.InputPeerChannel):
+            raise ValueError(f'The chat_id "{chat_id}" does not belong to a channel')
+
+        if period not in (0, 86400, 604800):
```
delivrance, line 65: I'd remove local checks like this as Telegram can allow more values at anytime. The user will get an up to date error straight from the server.

pyrogram/methods/chats/set_history_ttl.py
```diff
@@ -0,0 +1,83 @@
...
+
+                # Disabling
+                app.set_history_ttl("PyrogramChat", 0)
+
+        Raises:
+            ValueError: In case the chat_id doesn't belong to a supergroup or the time-to-live isn't 1 day or week.
+        """
+
+        peer = await self.resolve_peer(chat_id)
+        if not isinstance(peer, raw.types.InputPeerChannel):
```
delivrance, line 62: Same here. No checks like this as it became quite annoying copy pasting them all over places. The user will still get a meaningful and up to date error from the server.

pyrogram/methods/chats/set_history_ttl.py
```diff
@@ -0,0 +1,83 @@
...
+
+from typing import Union, Optional
+
+from pyrogram import raw
+from pyrogram import types
+from pyrogram.scaffold import Scaffold
+
+
+class SetHistoryTTL(Scaffold):
+    async def set_history_ttl(
```
delivrance, line 27: `set_chat_history_ttl` sounds better, so we can have `Chat.set_history_ttl` (I know we have get/iter_history and that should actually be get_chat_history and Chat.get_history for consistence).

pyrogram/types/messages_and_media/message.py
```diff
@@ -352,6 +352,7 @@ def __init__(
         voice_chat_started: "types.VoiceChatStarted" = None,
         voice_chat_ended: "types.VoiceChatEnded" = None,
         voice_chat_members_invited: "types.VoiceChatMembersInvited" = None,
+        history_ttl: int = None,
```
delivrance, line 355: Same as above. `chat_history_ttl` here.

pyrogram/types/messages_and_media/message.py
```diff
@@ -498,6 +501,9 @@ async def _parse(
             elif isinstance(action, raw.types.MessageActionInviteToGroupCall):
                 voice_chat_members_invited = types.VoiceChatMembersInvited._parse(client, action, users)
                 service_type = "voice_chat_members_invited"
+            elif isinstance(action, raw.types.MessageActionSetMessagesTTL):
+                history_ttl = action.period
+                service_type = "history_ttl"
```
delivrance, line 506: and here too, `"chat_history_ttl"`

delivrance (changes requested):

delivrance:
I'm also not fully convinced about keeping "TTL" in the convenience API, but at the moment I can't think of a better and short enough replacement.

pyrogram/methods/chats/set_history_ttl.py
delivrance: Plus, you can set the messages TTL on private chats too.

pyrogram/types/messages_and_media/message.py
```diff
@@ -498,6 +501,9 @@ async def _parse(
             elif isinstance(action, raw.types.MessageActionInviteToGroupCall):
                 voice_chat_members_invited = types.VoiceChatMembersInvited._parse(client, action, users)
                 service_type = "voice_chat_members_invited"
+            elif isinstance(action, raw.types.MessageActionSetMessagesTTL):
```
delivrance, line 504: Reading this made me think that maybe we should rename the whole thing to `set_chat_messages_ttl` or just `set_chat_ttl` since the feature applies to new messages not to the whole history (or can be seen as a global chat setting regardless of messages or history). What do you think? There's also a `ttl_period` field for messages (see the raw Message object) which I guess it will be set for all new messages when ttl is enabled.

delivrance: So, `set_chat_ttl()`, `Chat.set_ttl()`, `Message.chat_ttl_period`.

PR #683, opened 2021-05-14
Fix inline message id invalid

Ripeey:
It happened because it only returned the session of the first that it authorized to, even for other dc peers. So the fix was either to remove it completely or save and return sessions accordingly. There might be possibly a better clean way to do this.


delivrance:
> There might be possibly a better clean way to do this.

Definitely. Ideally the Client should be able to invoke methods on any DC at any point, not just in this context. I'm accepting this fix, though. Thanks.

PR #686, opened 2021-05-29
Added iterable download method and other features

jorgeajimenezl:
see #687


delivrance:
This feature, among with downloading into in-memory buffers, has been reworked recently and will be available in the next major release.

PR #690, opened 2021-06-02
Change variable name to avoid conflts with Telegram TL schema

allen0099:
Currently, the api compiler is using `data` as its function argument.  

However, [updateBotCallbackQuery](https://core.telegram.org/constructor/updateBotCallbackQuery) is use the same word `data` as its argument.  
It will cause the problem when the bot gets the callback query with no data (Like the game button).

Trying to fix issue #608 and make game button usable.

### DIFF
--- a/pyrogram/types/bots_and_keyboards/callback_query.py
+++ b/pyrogram/types/bots_and_keyboards/callback_query.py
@@ -110,7 +110,7 @@ class CallbackQuery(Object, Update):
         # ignoring/replacing errors, this way, button clicks will still work.
         try:
             data = callback_query.data.decode()
-        except UnicodeDecodeError:
+        except (UnicodeDecodeError, AttributeError):
             data = callback_query.data
 
         return CallbackQuery(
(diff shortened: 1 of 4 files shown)


pyrogram/types/bots_and_keyboards/callback_query.py
```diff
@@ -110,7 +110,7 @@ async def _parse(client, callback_query, users) -> "CallbackQuery":
         # ignoring/replacing errors, this way, button clicks will still work.
         try:
             data = callback_query.data.decode()
-        except UnicodeDecodeError:
+        except (UnicodeDecodeError, AttributeError):
```
delivrance, line 113: `AttributeError`, what for?

delivrance:
You forgot to rename some occurrences (I fixed those). Also, I've renamed to `b` (bytes, buffer) as it used to be.

pyrogram/types/bots_and_keyboards/callback_query.py
allen0099: Game button will return `data` with `None`

Sea-n:
Hey devs, any update on this PR?
I need this patch too. :bowing_man:

asimaranov:
I faced with this issue to. The following exception occurres if I click game button
IMG_20210812_122033_463. Please fix

PR #692, opened 2021-06-04
Unnecessary method call in get_media_group

Taijefx34:
Removed unnecessary method call


delivrance:
@Danipulok?

Danipulok:
Yup, Legenda's right. Don't know what was wrong with me when I left one unused, he told me about this before making current PR, he's right

However, I suggest moving comment 
`# There can be maximum 10 items in a media group.`
right before calling `messages = await self.get_messages(`

Taijefx34:
Closed due to conflict (out of date)
Pyro v 1.3.0

PR #693, opened 2021-06-05
Add KeyboardButtonPollType

SpEcHiDe:


pyrogram/types/bots_and_keyboards/keyboard_button_poll_type.py
```diff
@@ -0,0 +1,34 @@
...
+    
+    Parameters:
+        type (``str``):
+            If quiz is passed, the user will be allowed to create only polls in the quiz mode.
+            If regular is passed, only regular polls will be allowed.
+            Otherwise, the user will be allowed to create a poll of any type.
+    """
+    def __init__(
+        self,
+        type: str
```
delivrance, line 32: It seems like this can be optional, https://core.telegram.org/bots/api/#keyboardbuttonpolltype.

SpEcHiDe: I couldn't get the "regular" to work. (I don't know which raw type to use)
only quiz=True seems to work.

PR #695, opened 2021-06-11
Add update handlers for connected, disconnected and ready

alemidev:
Added
`@Client.on_ready()`
`@Client.on_connect()`
`@Client.on_disconnect()`
decorators (and related Handlers)

These events can be registered from smart plugins and can be grouped (unlike how `DisconnectHandler` was implemented).
To achieve this, "fake" updates are pushed on dispatcher's `updates_queue`. These are special "sentinel" classes defined in `pyrogram.session.session.py` and hold almost no information.

I had to map these handlers in the dispatcher. Also, the dispatcher used to just make a tuple of `parsed_update` (the object given out by the update parser). Since these updates hold no information to parse, they don't need any other argument in the callback other than `client`. To allow this, now `parsed_update` is checked, and if not iterable, a tuple is made. Like this, parsers can return iterables with multiple args (or none).

I'm not sure how docs are generated? I added the new methods and object inside the proper `.rst` files.

### DIFF
--- a/docs/source/api/decorators.rst
+++ b/docs/source/api/decorators.rst
@@ -49,7 +49,7 @@ Index
     - :meth:`~Client.on_poll`
     - :meth:`~Client.on_connect`
     - :meth:`~Client.on_disconnect`
-    - :meth:`~Client.on_client_ready`
+    - :meth:`~Client.on_ready`
     - :meth:`~Client.on_raw_update`
 
 -----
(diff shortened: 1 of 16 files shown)


alemidev:
Updated: there's only one handler + decorator now: `@Client.on_client_status`. There are filters to only register to certain events:
`@Client.on_client_status(filters.client_ready)` is `@Client.on_ready()`
`@Client.on_client_status(filters.client_disconnected)` is `@Client.on_disconnected()`

sam01101:
I recommend you squeeze all commits to one commit and force push for better looking :heart:

alemidev:
@sam01101 Commits can (and should) be squashed at merge time. I don't see why I should hide my progress (and mistakes!) from the PR itself

sam01101:
Oh really. Sorry for that.

delivrance:
Having handlers/decorators for such events is useful, but I don't like the implementation proposed here, especially because you merged them with the actual update dispatcher and filters.
I also prefer sticking with separate decorators because that's how it is now for all other decorators. Decorators should also have different names to reflect the method that trigger them (i.e.: start, stop); so, something like `on_starting` (before start), `on_started` (after start), `on_stopping` (before stop), `on_stopped` (after stop).

PR #706, opened 2021-06-26
Removed create_task

Mahesh0253:
No need to use loop.create_task in workers, because we awaiting immediately


delivrance:
Ok, the actual reason is that `create_task` placed there was spawning extra tasks and thus not staying within the pool and workers count limit of the method.

PR #712, opened 2021-07-06
New filter group_query

Princic-1837592:
Similar to filters.group but works but for Callback Queries


delivrance:
I believe the filter.group should be made to work with both kinds of events (see filters.regex for example) instead of creating a new filter.

PR #713, opened 2021-07-07
Bound method to stop_poll

New-dev0:


delivrance:
A poll is not a message per se, is a kind of media. This needs to go to the `Poll` class so you can do `Poll.stop()`.

PR #731, opened 2021-07-29
Recover gaps

Tenount:
@delivrance 
The changes will allow the client to process `raw.types.RpcResult` `updates.State` to fetching updates that arrived while the client was offline.
At least the missed messages. I checked it on echobot.

To enable set `recover_gaps=True`


drizzt:
It doesn't work with channels/supergroups

penn5:
It's not complicated enough to be correct.

septatrix:
I think recovering gaps should be the default - at least if a session already exists

PR #734, opened 2021-08-01
Display an error message to help solve the problem

hdcola:
If there is no error message when plugins are loaded, it will be very difficult for maintainers to know the problem and solve it.

Yesterday, I suddenly encountered the problem of not being able to load plugins, and later found out that a pip lib was not installed.

### DIFF
--- a/pyrogram/client.py
+++ b/pyrogram/client.py
@@ -738,7 +738,7 @@ class Client(Methods, Scaffold):
                     try:
                         module = import_module(module_path)
                     except ImportError:
-                        log.warning(f'[{self.session_name}] [LOAD] Ignoring non-existent module "{module_path}"')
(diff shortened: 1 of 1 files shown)


pyrogram/client.py
```diff
@@ -738,7 +738,7 @@ def load_plugins(self):
                     try:
                         module = import_module(module_path)
                     except ImportError:
-                        log.warning(f'[{self.session_name}] [LOAD] Ignoring non-existent module "{module_path}"')
+                        log.warning(f'[{self.session_name}] [LOAD] Ignoring non-existent module "{module_path}" error is "{e}"')
```
AndrielFR, line 741: Or just add the keyword `exc_info=True`.

PR #738, opened 2021-08-02
update the new and left chat member filters

idokar:
Filters for new and left chat member that supports the new update type


iiiiii1wepfj:
good

iiiiii1wepfj:
good but ChatMemberUpdated is not imported

idokar:
Corner case that wasn't catch

PR #739, opened 2021-08-07
Ability to run multiple bot with inline feature.

HasibulKabir:
Global session creates a problem for another bot. As that session was generated for another bot can't be used by this bot.


delivrance:
Thanks. I've change it so it uses the existing media session lock. This should also solve an issue about creating an async lock outside an async context (inside other threads, for example).

PR #741, opened 2021-08-10
update for the new and left chat member filters

idokar:
This update suports the new ChetMemberUpdate updare.

### DIFF
--- a/pyrogram/filters.py
+++ b/pyrogram/filters.py
@@ -471,14 +471,18 @@ channel = create(channel_filter)
 
 # region new_chat_members_filter
 async def new_chat_member_filter(_, __, m: Union[Message, ChatMemberUpdated]):
-    if isinstance(m, ChatMemberUpdated):
+    if isinstance(m, Message):
+        return bool(m.new_chat_members)
+    if m.new_chat_member:
+        threshold = 2
         ncm = m.new_chat_member
-        if ncm and ncm.status == 'member':
-            return bool(abs(m.date - ncm.joined_date) <= 3)
-        elif ncm and ncm.status == 'administrator' and not m.old_chat_member:
-            return bool(abs(m.date - ncm.joined_date) <= 3)
-        return False
-    return bool(m.new_chat_members)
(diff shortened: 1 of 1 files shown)


idokar:
This new functions have been checked on almost all the cases
All cases

pyrogram/filters.py
```diff
@@ -470,18 +470,35 @@ async def channel_filter(_, __, m: Message):
 # endregion
 
 # region new_chat_members_filter
-async def new_chat_members_filter(_, __, m: Message):
-    return bool(m.new_chat_members)
+async def new_chat_member_filter(_, __, m: Union[Message, ChatMemberUpdated]):
+    if isinstance(m, Message):
+        return bool(m.new_chat_members)
+    if m.new_chat_member:
+        threshold = 2
```
idokar, line 477: The best threshold is 2sec or 3sec

delivrance:
I believe we need separate filters (with new names) that don't mess with the existing ones (which are dedicated to Message objects because they are about events that generate the service message, whereas ChatMemberUpdated events are silent).
Any applicable situation can have a filter (i.e. joined, left, restricted, unrestricted, kicked/banned, unbanned, ...)

idokar:
> 
> 
> I believe we need separate filters (with new names) that don't mess with the existing ones (which are dedicated to Message objects because they are about events that generate the service message, whereas ChatMemberUpdated events are silent).
> Any applicable situation can have a filter (i.e. joined, left, restricted, unrestricted, kicked/banned, unbanned, ...)

Want me to upload it as a separate file with filters for ChatMemberUpdated or in the same file?

idokar:
> 
> 
> I believe we need separate filters (with new names) that don't mess with the existing ones (which are dedicated to Message objects because they are about events that generate the service message, whereas ChatMemberUpdated events are silent). Any applicable situation can have a filter (i.e. joined, left, restricted, unrestricted, kicked/banned, unbanned, ...)

@delivrance I would appreciate it if you could check out the new update

PR #744, opened 2021-08-12
improved async to sync wrapper

rking32:
Hi, Thanks for this awesome project. I have been using **Pyrogram** for a long time.
But recently got some issues with Threads + Event loops.

Here how i reproduced them.
https://gist.github.com/rking32/818bd60a6dd1e632ac61ba6f7e902a5a

Finally was able to fix them.

### DIFF
--- a/pyrogram/sync.py
+++ b/pyrogram/sync.py
@@ -41,6 +41,7 @@ def async_to_sync(obj, name):
             loop = asyncio.get_event_loop()
         except RuntimeError:
             loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
 
         if threading.current_thread() is threading.main_thread():
             if loop.is_running():
@@ -53,12 +54,13 @@ def async_to_sync(obj, name):
                     return loop.run_until_complete(consume_generator(coroutine))
         else:
             if inspect.iscoroutine(coroutine):
-                future = asyncio.run_coroutine_threadsafe(coroutine, main_loop)
-
                 if loop.is_running():
-                    return asyncio.wrap_future(future)
+                    async def coro_wrapper():
+                        return await asyncio.wrap_future(asyncio.run_coroutine_threadsafe(coroutine, main_loop))
+
+                    return coro_wrapper()
                 else:
-                    return future.result()
+                    return asyncio.run_coroutine_threadsafe(coroutine, main_loop).result()
 
             if inspect.isasyncgen(coroutine):
                 if loop.is_running():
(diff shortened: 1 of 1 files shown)


delivrance:
Thanks for the contribution. I have read through the code and was able to refactor it in order to improve readability and make it less cluttered. One trick was to create a new loop in the other, non-main thread; this allows to take into consideration two cases only, main & non-main threads. The examples provided still work fine with coroutines and generators.

pyrogram/sync.py
```diff
@@ -40,49 +40,31 @@ def async_to_sync_wrap(*args, **kwargs):
...
-                    async def coro_wrapper():
-                        return await asyncio.wrap_future(
-                            asyncio.run_coroutine_threadsafe(coroutine, main_loop))
+                future = asyncio.run_coroutine_threadsafe(coroutine, main_loop)
 
-                    return coro_wrapper()
-
-                return asyncio.run_coroutine_threadsafe(coroutine, main_loop).result()
+                if loop.is_running():
+                    return asyncio.wrap_future(future)
```
rking32, line 59: in here, is it ok to return just the future ? shouldn't be wrapped with a coro ?
```python
async def coro():
    return await asyncio.wrap_future(asyncio.run_coroutine_threadsafe(coroutine, main_loop))

return coro()
```
because this will submit request to the main loop and it will be called soon inside main loop
```python
future = asyncio.run_coroutine_threadsafe(coroutine, main_loop)
return asyncio.wrap_future(future)
```
example:
```python
async def main():
    fut = client.get_me()
    # suppose we will need to drive this method later
    # but in here get_me() method will be called soon before we explicitly await

    ...

    me = await fut
    # actually we need to do it here right ?
```

rking32 (commented):
much cleaner code, my code like a hell :D

pyrogram/sync.py
delivrance: > in here, is it ok to return just the future ? shouldn't be wrapped with a coro ?

Thanks for noticing. Correct, it should indeed, like you proposed initially. `run_coroutine_threadsafe` actually does the scheduling part and immediately starts the task underneath and we want that only in case of non-async code.

delivrance:
This should be fine now, however I'm a bit concerned about creating that new event loop.

rking32:
nice

pyrogram/sync.py
```diff
@@ -40,23 +40,33 @@ def async_to_sync_wrap(*args, **kwargs):
         try:
             loop = asyncio.get_event_loop()
         except RuntimeError:
-            loop = main_loop
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
```
delivrance, line 44: Adding `asyncio.set_event_loop(loop)` here fixed an issue where a new loop would be created every time a method was called in the same new thread.

rking32: yes. its a good point :)

Issue #698, opened 2021-06-16
AttributeError with ChatPermissions. ('ChatPermissions' object has no attribute 'can_send_stickers')


StarkGang:
## Checklist
- [x] I am sure the error is coming from Pyrogram's code and not elsewhere.
- [x] I have searched in the issue tracker for similar bug reports, including closed ones.
- [x] I ran `pip3 install -U https://github.com/pyrogram/pyrogram/archive/master.zip` and reproduced the issue using the latest development version.

## Description

After this [commit](https://github.com/pyrogram/pyrogram/commit/c9d0c5d75a7646b43cd7e8c1f45715fd1630f797), It seems that ```can_send_stickers``` has been removed. but it has been not updated [here](https://github.com/pyrogram/pyrogram/blob/4f585c156c1a2c6707793a8ad7f2f111515ca23b/pyrogram/methods/chats/restrict_chat_member.py#L82)

## Code

``` python
from pyrogram import Client, filters
from pyrogram.types import ChatPermissions

@Client.on_message(filters.me)
async def test(client, message):
    await client.restrict_chat_member(
            message.chat.id, message.reply_to_message.from_user.id, ChatPermissions(can_send_messages=False)
```

## Traceback

```
Traceback (most recent call last):
  File "<string>", line 4, in __aexec
  File "/usr/local/lib/python3.9/site-packages/pyrogram/methods/chats/restrict_chat_member.py", line 82, in restrict_chat_member
    send_stickers=True if not permissions.can_send_stickers else None,
AttributeError: 'ChatPermissions' object has no attribute 'can_send_stickers'
```

PR #746, opened 2021-08-19
Fix old non existing can_send_[sticker...] with can_send_other_messages

DatDraggy:
Fixes https://github.com/pyrogram/pyrogram/issues/698
Introduced by https://github.com/pyrogram/pyrogram/commit/c9d0c5d75a7646b43cd7e8c1f45715fd1630f797


delivrance:
Not really a bug, this was done on purpose to make permissions look like the ones in official clients for a reason: I noticed that if you set a more granular permissions with the method (I.e. deny gifs/stickers/games & allow inline bots) and open the chat settings with official clients (Android, tdesktop), the permissions will overwrite and will start to apply to all "other" media (I.e. in this example, inline bots become denied as well).

DatDraggy:
Yea I get that, but without this these types would result in `'ChatPermissions' object has no attribute 'can_send_stickers'` because my changes weren't being made. `can_send_stickers` doesn't exist, only `can_send_other_messages` is a thing now.

DatDraggy:
Still relevant. Please take a look at the diff. This implements what you're describing.

PR #750, opened 2021-09-03
hotfix for iter_dialogs()

holnburger:
Telegram now always seems to return at least one dialog with the offset_date set to the oldest dialog. A hotfix was implemented to prevent endless loops.


jonatan1609:
Why `<=`? What if there is only one (1) dialog? Why would you `return` in such a case?. You probably want `<`, which will be the same as `if len(dialogs) == 0` (since len cannot be negative) but then such a statement is equal to `if not dialogs`.
Useless PR.

User087:
A related issue #756

PR #758, opened 2021-09-21
Added missing param. `schedule_date` in message.py

sh-himanshu:


StarkBotsIndustries:
Bound methods should accept `schedule_date` parameter as a `datetime` object, not `int`

PR #764, opened 2021-10-02
Add: a new parameter `input_function` to Client

jonatan1609:
The idea is to allow users to give input without having to type it in the console, but prepared inputs.
The function should be defined as follows and always must return a string which will be passed as the input.

`input_type` is the required input.
`formats` is a tuple containing the strings that should be formatted into the original prompt (if needed).
`hide` determines whether the input should be hidden while is being typed or not (in case it's "secret" like entering 2FA password).

```python
async def your_function(
    input_type: Scaffold.InputType,
    formats: Sequence[str] = (),
    hide: bool = False
) -> str: ...
```

For example, their input could be brought this way:

```python
async def your_function(
    input_type: Scaffold.InputType,
    formats: Sequence[str] = (),
    hide: bool = False
) -> str:
    if input_type == Scaffold.InputType.EnterPhoneOrToken:
        return PHONE_NUMBER
  
    # Make sure to handle all inputs correctly.
```

You have to handle all types of inputs so the function could always return a corresponding string.
You can also add a default `input` function with the original prompt:

```python
else:
    return await pyrogram.utils.ainput(input_type.value.format(*formats))
```

### DIFF
--- a/pyrogram/client.py
+++ b/pyrogram/client.py
@@ -202,7 +202,7 @@ class Client(Methods, Scaffold):
 
                 hide (``bool``, *optional*):
                     Determines whether the input should be hidden while is being typed.
-                    Determined by the `hide_password` parameter of ``~pyrogram.Client`.
+                    Determined by the `hide_password` parameter of ``~pyrogram.Client``.
                     Default is False.
             Example:
                 ..code-block:: python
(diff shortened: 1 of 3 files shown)


delivrance:
Something like this once existed but was later removed because the implementation was quite messy to maintain, document and use. Plus, I believe logins should be handled with a step-by-step logic using the relevant methods instead, where you can handle each step the way you like instead of being bound to pre-defined callbacks.

PR #767, opened 2021-10-11
save_file: allow async read

mayli:


delivrance:
That's not how it works, and it's not that simple.

PR #768, opened 2021-10-11
Add search_global_count and search_messages_count methods

delivrance:


zeroone2numeral2:
Since the limit is hard-coded to 1, does it really make sense to return `len(r.messages)` if for some reason the result doesn't have a `count` property, instead of raising an exception?

PR #770, opened 2021-10-21
Add `video_start_ts` parameter to `set_chat_photo`

roj1512:


delivrance:
Thanks, I've updated the description and made the parameter hint float only as it is enough for integers too.

roj1512:
Io sono love you, thanks as f!

PR #806, opened 2021-12-06
Change MAX_USER_ID to integer 64 limit

sam01101:
`999999999999` is not enough for the telegram user id, perhaps int64 is the official limit of the ID.


delivrance:
Hi. That constant was taken from [here](https://github.com/tdlib/td/commit/02ae56ae8a209755134efdf1c6c41d1769991e61).

delivrance:
Now that I notice, MAX_USER_ID should be `2^40 - 1` according to that commit, which is the closest power of two to `999999999999`

sam01101:
oh.

PR #809, opened 2021-12-08
Added admin filter

belarusaed:
Hello, I made admin filter for chats, may be it will be useful for anybody.

### DIFF
--- a/pyrogram/filters.py
+++ b/pyrogram/filters.py
@@ -943,3 +943,14 @@ async def dan_filter(_, __, m: Message):
 
 dan = create(dan_filter)
 # endregion
+
+
+# region admin_filter
+async def admin_filter(_, __, message: Message):
+    return (message.sender_chat and message.sender_chat.id == message.chat.id) or \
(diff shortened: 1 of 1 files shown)


zeroone2numeral2:
every time an update is tested against the filter, a get_chat_member request is executed. This is a network-bound implicit behavior that might result in unexpected results (eg. being rate-limited by the server), I believe it should be up to the developer whether to use (and how to implement) such filters

belarusaed:
> every time an update is tested against this filter, a `Client.get_chat_member()` request is executed

Yes, but this is the only way to know that the user is an administrator, as I know. If someone wants to check the admin, they still have to repeat it, and this, it seems to me, is not a rare use case, therefore it can be useful as a standard filter. 

But if this is really such a bad practice, then I am by no means insisting.

nocturn9x:
Honestly I doubt this is of any usefulness. Most use cases for this are when writing chat manager bots, which need specific behavior tuned for their use case. A generic filter is rarely useful

Issue #664, opened 2021-04-11
client.download_media consistently fails for photos for _some_ messages -503 Timeout


ChrisRxyz:
## Checklist
- [X] I am sure the error is coming from Pyrogram's code and not elsewhere.
- [X] I have searched in the issue tracker for similar bug reports, including closed ones.
- [X] I ran `pip3 install -U https://github.com/pyrogram/pyrogram/archive/master.zip` and reproduced the issue using the latest development version.

## Description
Im using pyrogram to forward messages to another system, using a raw handler to process the messages and also download corresponding media (mostly images).
Im using the client.download_media method. This works for almost all messages but for some it just wont download the images and return None as a path. I could not figure out why it happens but identified some messages for which it consistently happens. 

I made a small testcase that fetches one of these messages from a channel and tries to download the media.

The message looks like this:

```
{
    "_": "Message",
    "message_id": 5869,
    "sender_chat": {
        "_": "Chat",
        "id": -1001133587448,
        "type": "channel",
        "is_verified": false,
        "is_restricted": false,
        "is_creator": false,
        "is_scam": false,
        "is_fake": false,
        "title": "Klondike (Crypto Rush) ⚠️",
        "username": "CryptoKlondike",
        "photo": {
            "_": "ChatPhoto",
            "small_file_id": "AQADBAATtJV9J10AAwIAAwjEyecW____ZkFiWLn_WDe_GwQAAR4E",
            "small_photo_unique_id": "AQADtJV9J10AA78bBAAB",
            "big_file_id": "AQADBAATtJV9J10AAwMAAwjEyecW____ZkFiWLn_WDfBGwQAAR4E",
            "big_photo_unique_id": "AQADtJV9J10AA8EbBAAB"
        },
        "dc_id": 4
    },
    "date": "2021-02-19 10:35:42",
    "chat": {
        "_": "Chat",
        "id": -1001133587448,
        "type": "channel",
        "is_verified": false,
        "is_restricted": false,
        "is_creator": false,
        "is_scam": false,
        "is_fake": false,
        "title": "Klondike (Crypto Rush) ⚠️",
        "username": "CryptoKlondike",
        "photo": {
            "_": "ChatPhoto",
            "small_file_id": "AQADBAATtJV9J10AAwIAAwjEyecW____ZkFiWLn_WDe_GwQAAR4E",
            "small_photo_unique_id": "AQADtJV9J10AA78bBAAB",
[...]

PR #784, opened 2021-11-22
Fixed 503 error when downloading files

JohnSmith09512:
https://github.com/pyrogram/pyrogram/issues/664

PR #812, opened 2021-12-10
fix for client.download_media consistently fails for photos for big channels -503 Timeout #664

AduchiMergen:
some times server don't send response if offset more then file size (i think some file storages don't support this behavior)

so fix in one line: need change condition for break loop to len(chunk) < limit

### DIFF
--- a/pyrogram/client.py
+++ b/pyrogram/client.py
@@ -916,9 +916,6 @@ class Client(Methods, Scaffold):
                     while True:
                         chunk = r.bytes
 
-                        if not chunk:
-                            break
-
                         f.write(chunk)
 
                         offset += limit
@@ -938,6 +935,9 @@ class Client(Methods, Scaffold):
                             else:
                                 await self.loop.run_in_executor(self.executor, func)
 
+                        if len(chunk) < limit:
+                            break        
+                        
                         r = await session.send(
                             raw.functions.upload.GetFile(
                                 location=location,


MyNameIsDark01:
Already exists a fix for this issue #784

AduchiMergen:
@MyNameIsDark01 fix in #784 have two problems:
1. change limit from 1MB to 32 KB
2. don't save last chunk of file, so files corrupted

JohnSmith09512:
> @MyNameIsDark01 fix in #784 have problems:
> 
>     1. change limit from 1MB to 32 KB
> 
>     2. don't save last chunk of file, so files corrupted
> 
>     3. then we download file by file_id we don't know file_size, so file_size var = 0  and file not downloaded at all

Use @AduchiMergen fix instead, i posted mine in a rush just to hint at the problem and to use in a personal project.

I reversed source code of the original telegram app trying to figure out this problem and took chunk_size from it (32KB constant)

delivrance (approved):

righteon:
Thanks for (hopfully) finally fixing this. 

 I have pointed out this bug **April 11th**, provided a concise testcase to confirm the bug, I even **donated money** to the project to get it fixed, drew attention to the bug here and in the telegram channel but still was always brushed off and told it was a "server problem" and there was nothing there could be done. VERY, very dissapointing - to say the least!
 
 In the meantime I had to switch to another python telegram framework and re-implement my logic to proceed with my work.
 
 **Next time, please actually try to run the testcases that people provide and take bug reports seriously. Thanks..** Otherwise this is a great project.

PR #860, opened 2022-01-07
Remove unnecessary method call in `get_media_group`

Danipulok:
**Remove unnecessary method call in `get_media_group`**

Since currently `get_media_group` method is in stable version, also as `copy_media_groups`, which uses `get_media_group`, it's even more critical now. Currently, we use `get_messages` method twice, when only one call is needed.

[Old](https://github.com/pyrogram/pyrogram/pull/692) pull request from @Taijefx34 is closed because of conflicts, so this is a new one.

@delivrance, I really hope you could merge this as soon as possible and maybe even add it to hot fix version v1.3.2.

### DIFF
--- a/pyrogram/methods/messages/get_media_group.py
+++ b/pyrogram/methods/messages/get_media_group.py
@@ -51,13 +51,10 @@ class GetMediaGroup(Scaffold):
                 In case target message doesn't belong to a media group.
         """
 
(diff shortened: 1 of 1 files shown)


Danipulok:
@delivrance ?

This should be fixes ASAP, please.

PR #862, opened 2022-01-07
Fix spoiler html unparsing

alissonlauffer:
- The current spoiler implementaion unparses both strikethrough and spoiler tags with \<s\>, making them indistinguishable:


A9FM:
+

PR #867, opened 2022-01-10
Badshah346

Badshah653:
from pyrogram import Client, filters

app = Client("my_account")

@app.on_message(filters.private)
async def hello(client, message):
    await message.reply("Hello from Pyrogram!")

app.run()

### DIFF
--- a/Badshah
+++ b/Badshah
@@ -1,0 +1,11 @@
+from pyrogram import Client, filters
+
+app = Client("my_account")
+
+
+@app.on_message(filters.private)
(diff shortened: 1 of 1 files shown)


Badshah653:
Yes

Badshah653:
from pyrogram import Client, filters

app = Client("my_account")


@app.on_message(filters.private)
async def hello(client, message):
    await message.reply("Hello from Pyrogram!")


app.run()

Badshah653:
Th

Issue #870, opened 2022-01-11
html unparse produces wrong results if two entities are close


Princic-1837592:
## Checklist
- [x] I am sure the error is coming from Pyrogram's code and not elsewhere.
- [x] I have searched in the issue tracker for similar bug reports, including closed ones.
- [x] I ran `pip3 install -U https://github.com/pyrogram/pyrogram/archive/master.zip` and reproduced the issue using the latest development version.

## Description
When calling ```message.text.html``` I get a wrong result in case two entities are consecutive

``` python
msg = await client.send_message(chat_id, '<code>entity 1</code><code>entity 2</code>')
print(msg.text.html)
# expected: <code>entity 1</code><code>entity 2</code>
# actual result: <code>entity 1<code></code>entity 2</code>
```
This happens because at line 179 of ```html.py``` the list of entities is sorted by offset, but two consecutive entities have same offset: one ends and the other starts at the same point in the string so they are sorted as equal offset value, but have different tag (one closing and one opening)

## Traceback
No traceback raised, just wrong html format

Princic-1837592:
Fixed in this pull request https://github.com/pyrogram/pyrogram/pull/872

delivrance:
Fixed in #885. Discussion at #872.

PR #872, opened 2022-01-12
Fixed error while unparsing consecutive entities

Princic-1837592:
As shown here https://github.com/pyrogram/pyrogram/issues/870
Now sorting also takes in consideration the original order of the entities


delivrance:
I found that a much simpler solution would be using `entities_offsets.reverse()` or iterate directly in reverse order instead of an intricate composition of `map/sorted/enumerate` as entities are already coming sorted as wanted from the server. The code just needs the reverse.

Princic-1837592:
But we only need to reverse entities with same offset, not the entire list. This is what enumerate is for

delivrance:
What I'm saying is that the code builds the unparsed string from end to start and since entities are already sorted by the server we simply need to iterate through `reversed(entities_offsets)` in the loop below.

Princic-1837592:
You are right. So the "problem" is the stability of `.sort`

delivrance:
Why closing it? You can edit this PR.

Princic-1837592:
Sorry, I'm not so used to github

delivrance:
You should create one branch per PR. I see you included other changes in this one.

Princic-1837592:
I wanted to add a commit to this PR https://github.com/pyrogram/pyrogram/pull/712 but the commit was automatically added to this one. Maybe the other one is too old

delivrance:
Either cleanup your master branch to include only the relevant changes or create a new branch, apply changes there and open a new PR.

PR #885, opened 2022-01-29
Fixed error while unparsing consecutive entities

Princic-1837592:


delivrance:
Original discussion at #872.

PR #886, opened 2022-01-29
Now filters.group supports callback queries

Princic-1837592:


pyrogram/filters.py
```diff
@@ -448,7 +448,8 @@ async def private_filter(_, __, m: Message):
 # endregion
 
 # region group_filter
-async def group_filter(_, __, m: Message):
+async def group_filter(_, __, u: Update):
+    m = u.message if isinstance(u, CallbackQuery) else u
```
AndrielFR, line 452: But what if the callback is a `via_bot` one?

Princic-1837592: The filter fails, as it did when applied to a callback query

AndrielFR: sad

delivrance:
We should extend the same enhancement to all relevant filters (private, channel, others?) and use an approach similar to `filters.regex` (multiple cases and raising an error for non-applicable updates). Plus, the `message` attribute for callback queries can be None; this needs to be handled.

PR #892, opened 2022-02-03
Now unparses overlapped entities

Princic-1837592:
Simple reversed method didn't work with overlapping entities

### DIFF
--- a/pyrogram/parser/html.py
+++ b/pyrogram/parser/html.py
@@ -174,8 +174,17 @@ class HTML:
 
             entities_offsets.append((start_tag, start,))
             entities_offsets.append((end_tag, end,))
+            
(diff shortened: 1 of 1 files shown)


Princic-1837592:
Sorry to keep this topic alive but the `reversed` method solved the problem with consecutive entities but introduced another problem with overlapping entities. Example code [here](https://t.me/pyrogramchat/397594). Original discussion [here](https://github.com/pyrogram/pyrogram/pull/885)

delivrance:
Indeed, relying on server's response order was breaking the parser in other scenarios. Your first proposal is a safer solution (i.e. explicitly sorting by offset as it used to be, plus original position). Thanks.

PR #901, opened 2022-02-13
fix repr issue

nikisalli:


gustch:
This change will result in an error: AttributeError 'int' object has no attribute 'QUALNAME'. I don't why it is.But when i change it back(api), it works.

PR #902, opened 2022-02-13
run all sqlite operations on executor

andrew-ld:


andrew-ld:
I noticed after doing this pr that sometimes executemany fails with "sqlite3.InterfaceError: Error binding parameter 0 - probably unsupported type.", I honestly don't know why this happens.

delivrance:
Not sure what you are trying to do here, but you latest change basically reverts to what it originally was before the PR. For some reason you replaced that query in your PR. For some other reasons you decided to ignore/replace context managers. I don't think I'm going to accept this.

andrew-ld:
@delivrance no this is not the problem, looking at the stacktrace is from internal pyrogram code, https://github.com/pyrogram/pyrogram/blob/7edfda7f417c661344eac11e7a09613cc882afcf/pyrogram/methods/advanced/send.py#L84 

I had added the for as a stupid initial workaround, now making list(peers) seems to work with executemany without a problem.

andrew-ld:
>  For some other reasons you decided to ignore/replace context managers. I don't think I'm going to accept this.

are you talking about the removal of inspect in _accessor method?

delivrance:
No, I was talking about the removal/replacement of some of the context managers which may lead to unwanted behaviours. This needs a more thorough review.

andrew-ld:
@delivrance the only one that needs to be checked might be the context manager of the concurrency lock which should be replaced with an asynchronous lock if it is still needed, the rest of the context managers have just been moved.

andrew-ld:
I remind you that since I use a threadpool with size limited to 1 all queries run in series on the same thread, many of the cases where there were locks are no longer necessary, in fact sqlite is now only accessed by a single thread and there is no need to use check_same_thread=False.

pyrogram/storage/file_storage.py
```diff
@@ -1,5 +1,5 @@
 #  Pyrogram - Telegram MTProto API Client Library for Python
-#  Copyright (C) 2017-present Dan <https://github.com/delivrance>
+#  Copyright (C) 2017-2021 Dan <https://github.com/delivrance>
```
AndrielFR, line 2: 2021?

AndrielFR:
why u don't use `aiosqlite`?

pyrogram/storage/file_storage.py
andrew-ld: I need to merge the master branch to update the changes

andrew-ld:
I did not want to add a dependency, if dan approves I can switch to aiosqlite

AndrielFR:
> I did not want to add a dependency, if dan approves I can switch to aiosqlite

oh, cool

pyrogram/storage/file_storage.py
AndrielFR: i see

PR #907, opened 2022-02-20
Invert storage loading order (for mypy)

alemidev:
mypy assumed Client.storage is of type MemoryStorage because the first time self.storage gets assigned (in code), it's of MemoryStorage type.

Move the generic Storage assignation before. Has no effect on behavior, but makes mypy know Client.storage is of type Storage.
As bonus points, makes even more clear for readers that you can pass a custom Storage type as name.

(putting `storage : Storage` in Client definition would also solve this)


delivrance:
At this point I think there should be a dedicated argument `session_string`, like it exists for `bot_token` so that we don't pollute the `session_name`.

PR #937, opened 2022-03-15
Bound method `react` to send reaction

Krishna-Singhal:


Krishna-Singhal:
@delivrance any review or updates ?

PR #964, opened 2022-04-24
download media on windows fix

LORD-ME-CODE:
mmmmmmmm


delivrance:
Explanation? This will likely break elsewhere.

LORD-ME-CODE:
> Explanation? This will likely break elsewhere.

No, (in_memory) and (not in_memory) are in different cases, so this should not throw an error.
But you can make an (try...except) block for this line if you are not sure

delivrance:
Ok, the `try...except` block might be a leftover after testing `NamedTemporaryFile` with automatic deletion, but since the temporary file settings used don't allow the automatic deletion of the file on close() this change makes sense because the file is moved from the temp directory to the final download directory.

PR #966, opened 2022-04-25
added timezone aware object

SmartManoj:


delivrance (approved):

delivrance:
I've reverted this because it's causing issues when using `datetime.now() - message.date` (`TypeError: can't subtract offset-naive and offset-aware datetimes`). Maybe you should have explained beforehand why this change was needed.

SmartManoj:
One should use `datetime.now().astimezone()`

SmartManoj (commented):
[REDACTED]

pyrogram/utils.py
```diff
@@ -53,8 +53,7 @@ def get_input_media_from_file_id(
     file_type = decoded.file_type
 
     if expected_file_type is not None and file_type != expected_file_type:
-        raise ValueError(
-            f"Expected {expected_file_type.name}, got {file_type.name} file id instead")
+        raise ValueError(f"Expected {expected_file_type.name}, got {file_type.name} file id instead")
```
SmartManoj, line 56: What's the code width the project follows?

delivrance:
I much prefer to allow simpler usages such as `datetime.now() + timedelta(...)` or `datetime(y, m, d, ...)` instead of having to deal with timezones. This seems to work without any issue because datetimes are only used as interface for unix timestamps, they only need to represent them.

pyrogram/utils.py
delivrance: 120 characters

PR #978, opened 2022-04-28
Allow to use custom storages

mikterbom:
Please allow the use of custom storages. I use it, for example, to restore non-truncated with peer info sessions from my own SQL dump.


drizzt:
Please fix this PR, but I like the idea

PR #982, opened 2022-04-30
Added missing edited filter

Desiders:


AlbertEinsteinTG (commented):
There is no need of this filter anymore. You have seperate decorator for edited messages. Therefore you won't receive edited messages in on_message decorator.

PR #1036, opened 2022-07-10
Fix inline quirks and mistakes

DavideGalilei:
Inline mode done right:
- Fixed `FilePartMissing: Telegram says: [400 FILE_PART_X_MISSING] - Part 0 of the file is missing from storage (caused by "messages.EditInlineBotMessage")` when editing a media in a chat which DC id is different by the bot's one
- `inline_session.get_session` always assumed to be in production server, changed it to `client.storage.test_mode()`
- `force_file=True` for `InputMediaDocument`
- `nosound_video=True` for `InputMediaAnimation`
- Implemented retries due to a server race condition which might lead to `MEDIA_EMPTY`

### DIFF
--- a/pyrogram/methods/messages/edit_inline_media.py
+++ b/pyrogram/methods/messages/edit_inline_media.py
@@ -18,16 +18,20 @@
 
 import os
 import re
+import asyncio
 
 import pyrogram
 from pyrogram import raw
 from pyrogram import types
 from pyrogram import utils
+from pyrogram.errors import RPCError, MediaEmpty
 from pyrogram.file_id import FileType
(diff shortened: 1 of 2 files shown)


DavideGalilei:
This example works correctly now:
```python
...

@bot.on_callback_query()
async def audio(bot: Client, query: CallbackQuery):
    await query.edit_message_media(
        media=InputMediaAudio("test.mp3"),
    )

...
```

DavideGalilei:
Note: I tested all cases with Video, Photo, Audio, Animation, Document, Web Document, and they all work as expected.

PR #1041, opened 2022-07-15
Filter out empty entities internally

Harsh-br0:
I guess it's fine to handle empty entities internally to avoid ENTITY_BOUNDS_INVALID , so the client won't send the empty entities

### DIFF
--- a/pyrogram/parser/html.py
+++ b/pyrogram/parser/html.py
@@ -140,6 +140,8 @@ class HTML:
 
             entities.append(entity)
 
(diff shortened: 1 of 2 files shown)


Harsh-br0:
> Sure, but you need to apply changes to the `parse` method in `parser/html.py` instead. I suggest to add a line that overwrites the entities, i.e.: `entities = list(filter...`, before the `return` happens to avoid a too messy one-liner.

Alright, applied the required changes to `parser/html.py `
Also the formatter did some useless changes too so please don't mind them

PR #1059, opened 2022-08-11
add min_id & max_id support in [messages.getHistory]

kkwilllo:
According to the official document (https://core.telegram.org/method/messages.getHistory)

max_id: Can be used to only return results with ID strictly smaller than max_id (e.g. message ID)
min_id: Can be used to only return results with ID strictly greater than min_id(e.g. message ID)

so added these parameters accordingly


Kacnep89:
This is a very necessary function. Why hasn't it been added yet?

kkwilllo:
> This is a very necessary functionality. Why hasn't it been added yet?

@delivrance

RaphxelS:
would really like this too

PR #1066, opened 2022-08-15
Fix get_custom_emoji_stickers Int type

omg-xtao:
Although it says in the function's comment that an int can be passed directly, you should actually pass a list.

This PR is used to allow the function to be called when parameters of type Iterable and int are provided.


delivrance:
The docstring was misleading and needs to be fixed. The rest is ok. `List[int]` is the accepted type.

delivrance:
The code *before* all the changes proposed in this PR is ok except for one line in the docstrings (which I fixed). There's no need to change anything else.

PR #1080, opened 2022-08-30
Added tempfile deletion in case of asyncio task cancellation.

el-garro:
Note: The asyncio.CancelledError() is re-raised to propagate to the caller


el-garro:
I'm and idiot, I thought had to close it

PR #1118, opened 2022-10-15
added `language` to entities when unparsing

Princic-1837592:
Added the field `language` when unparsing `PRE` entities (both markdown and html), which was being completely ignored.
Not added when parsing (yet)


delivrance:
I have tidied up the checks and also added a way to parse language codes via markdown.
The "pre" tag parsing and unparsing now works for both html and markdown.

PR #1131, opened 2022-10-31
Fix TypeError in method edit_message_media

Neon4o4:
Fix "TypeError: expected str, bytes or os.PathLike object, not BytesIO" when passing an InputMedia with .media attribute being a BytesIO to method edit_message_media

test case

```python
...

client = Client(**config)
with client:
    doc_io = io.BytesIO(b'aabb123')
    doc_io.name = 'media'
    msg: pyrogram.types.Message = client.send_document(chat_id=test_channel, document=doc_io)
    doc_io.seek(0)
    msg.edit_media(InputMediaDocument(media=doc_io))  # <- got "TypeError: expected str, bytes or os.PathLike object, not BytesIO", fixed in this pr
```

### DIFF
--- a/pyrogram/methods/messages/edit_message_media.py
+++ b/pyrogram/methods/messages/edit_message_media.py
@@ -91,6 +91,9 @@ class EditMessageMedia:
         if caption is not None:
             message, entities = (await self.parser.parse(caption, parse_mode)).values()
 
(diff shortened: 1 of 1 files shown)


TrixiS:
Looks like this PR needs to be merged. The issue is still happening

PR #1134, opened 2022-11-06
Fix on_disconnect decorator not working

01101sam:
Believe or not, this is the only handler that's not working

```py3
@Client.on_disconnect()
async def disconnect(client: Client):
    pass
````


01101sam:
Still, Dan is not going to merge this until next year :/

delivrance:
🐈

PR #1153, opened 2022-12-05
Chore: Adds support for web page preview the right way

ruizlenato:
**With WebPreview:**
image

**Without WebPreview:**
image

### DIFF
--- a/pyrogram/types/messages_and_media/message.py
+++ b/pyrogram/types/messages_and_media/message.py
@@ -254,7 +254,7 @@ class Message(Object, Update):
 
         views (``int``, *optional*):
             Channel post views.
-	    
+
 	forwards (``int``, *optional*):
             Channel post forwards.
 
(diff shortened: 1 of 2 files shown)


delivrance:
How is removing all attributes "the right way"?

ruizlenato:
> How is removing all attributes "the right way"?

Because the current methods don't return anything and telegram doesn't tell you if the media inside the WebPreview is Image or Document, it only returns the method and the WebPreview ID 

Ex:
 ```
   "web_page": {
        "_": "WebPage",
        "id": "3964102179771624742"
    },
```

delivrance:
The web page might not be available immediately because it takes some time to be generated by the server and until then you have to wait and re-fetch the message. There probably is a raw update to signal when this happens for a given message.

ruizlenato:
> The web page might not be available immediately because it takes some time to be generated by the server and until then you have to wait and re-fetch the message. There probably is a raw update to signal when this happens for a given message.

I have sent the same link more than 10 times within 5 minutes and nothing, Pyrogram currently does not return that there is a media in the message so I did this PR. Telegram currently only returns the WebPage ID.

In the Pull Request message I showed you how Pyrogram currently behaves and what it looks like with my pr

delivrance:
> Pyrogram currently does not return that there is a media in the message so I did this PR

The PR was invalid nonetheless. The current behavior of Pyrogram is to show fully available web pages (which are likely restricted server-side to user accounts only) and discard empty or pending ones.

PR #1162, opened 2022-12-12
Handle all given updates, avoid short circuit

### LINKED ISSUE #1161: `Client.handle_updates`: `is_min` looks very strange
### Checklist

- [X] I am sure the error is coming from Pyrogram's code and not elsewhere
- [X] I have searched in the issue tracker for similar bug reports, including closed ones
- [X] I ran `pip3 install -U https://github.com/pyrogram/pyrogram/archive/master.zip` and reproduced the issue using the latest development version

### Description

PR coming soon. [There](https://github.com/pyrogram/pyrogram/blob/master/pyrogram/client.py#L488) is a very strange place:
```python
    async def handle_updates(self, updates):
        if isinstance(updates, (raw.types.Updates, raw.types.UpdatesCombined)):
            is_min = (await self.fetch_peers(updates.users)) or (await self.fetch_peers(updates.chats))
```

If `updates.users` has `is_min`, `updates.chats` will not be processed

### Steps to reproduce

None

### Code example

```python
>>> def t(name):
...     async def f(res):
...         print(name)
...         return res
...     return f
... 
>>> t1 = t("t1")
>>> t2 = t("t2")
>>> 
>>> await t1(True)
t1
True
>>> (await t1(True)) or (await t2(True))
t1
True
>>>
```

### Logs

_No response_

kai3341:
Fixes #1161

### DIFF
--- a/pyrogram/client.py
+++ b/pyrogram/client.py
@@ -485,7 +485,10 @@ class Client(Methods):
 
     async def handle_updates(self, updates):
         if isinstance(updates, (raw.types.Updates, raw.types.UpdatesCombined)):
-            is_min = (await self.fetch_peers(updates.users)) or (await self.fetch_peers(updates.chats))
+            is_min = any((
+                await self.fetch_peers(updates.users),
+                await self.fetch_peers(updates.chats),
+            ))
 
             users = {u.id: u for u in updates.users}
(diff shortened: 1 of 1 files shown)


delivrance:
Thanks for spotting and fixing this issue!

PR #1171, opened 2022-12-18
Refactor at `Client.fetch_peers` extraction `channel_id`

kai3341:
It's too difficult to read this code. I simplified if

### DIFF
--- a/pyrogram/client.py
+++ b/pyrogram/client.py
@@ -483,6 +483,34 @@ class Client(Methods):
 
         return is_min
 
+    def __extract_channel_id__from_message(self, update):
+        """
+        Re-implementation extraction `channel_id` logic as is
+        Here we trying to extract it from message
+        """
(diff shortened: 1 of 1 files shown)


kai3341:
It looks it's first step of refactoring. Now it's easy to see:
-- `channel_id` at `Client.fetch_peers` uses only in one case but calculates always
-- instead of `getattr(..., ..., None) or getattr(..., ..., None)` it's good idea to use `getattr(..., ..., getattr(..., ..., None))`
-- maybe invert `getattr` action order
-- maybe extract attr via type checking

kai3341:
I found [patchy](https://pypi.org/project/patchy/) so I'm solving my issues
@delivrance you may close this PR. It change nothing in behavior

kai3341:
I have some ideas, so I'll implement it, push it and then remove `WIP` status

Issue #1172, opened 2022-12-18
Possible memory leakage in /types/object.py:29


xk4rim:
### Checklist

- [X] I am sure the error is coming from Pyrogram's code and not elsewhere
- [X] I have searched in the issue tracker for similar bug reports, including closed ones
- [X] I ran `pip3 install -U https://github.com/pyrogram/pyrogram/archive/master.zip` and reproduced the issue using the latest development version

### Description

I've a Telegram bot that's used by a few users. I just noticed that my virtual machine restarts every week due to its memory being full. Looking at the chart below, this is clearly caused by a memory leak:

![image](https://user-images.githubusercontent.com/49250671/208301028-5a6b1669-2f4b-452f-9716-0b44b00d1fde.png)

I've collected these three memory snapshots which are approximately ten commands apart:

```
Top 10 lines
#1: /usr/lib/python3.10/linecache.py:137: 392.3 KiB
    lines = fp.readlines()
#2: /usr/lib/python3.10/posixpath.py:372: 353.6 KiB
    path = sep*initial_slashes + path
#3: /usr/lib/python3.10/inspect.py:877: 72.0 KiB
    modulesbyfile[f] = modulesbyfile[
#4: /usr/lib/python3.10/inspect.py:874: 72.0 KiB
    _filesbymodname[modname] = f
#5: /home/USER/.local/lib/python3.10/site-packages/pyrogram/types/object.py:29: 17.1 KiB
    self._client = client
#6: /home/USER/.local/lib/python3.10/site-packages/pyrogram/raw/core/tl_object.py:33: 13.4 KiB
    return cast(TLObject, objects[int.from_bytes(b.read(4), "little")]).read(b, *args)
#7: /usr/lib/python3.10/asyncio/events.py:80: 5.0 KiB
    self._context.run(self._callback, *self._args)
#8: /home/USER/.local/lib/python3.10/site-packages/pyrogram/storage/sqlite_storage.py:135: 4.3 KiB
    self.conn.executemany(
#9: /home/USER/.local/lib/python3.10/site-packages/pyrogram/sync.py:53: 4.1 KiB
    coroutine = function(*args, **kwargs)
#10: /home/USER/.local/lib/python3.10/site-packages/pyrogram/raw/core/primitives/int.py:30: 3.7 KiB
    return int.from_bytes(data.read(cls.SIZE), "little", signed=signed)
584 other: 284.1 KiB
Total allocated size: 1221.7 KiB
```

```
[...]

PR #1176, opened 2022-12-20
Control the number of cached messages via message_cache in Client

xk4rim:
This commit adds the optional parameter message_cache to Client to control the number of cached messages.
This can significantly help low-end machines control their memory usage.

See #1172.


Mahesh0253:
Small suggestion, use `message_cache_size` name for variable instead of `message_cache` as same name is already used to store message cache instance.

PR #1195, opened 2023-01-12
Added ability to set the maximum amount of concurrent transmissions

CyanBook:


delivrance:
Thanks for the proposal, but a different solution has been chosen instead. See 82b8c7792e028a7ea5f855f2e7e1fe13bdabbfcb.

PR #1203, opened 2023-01-26
Add support for max_id and min_id

rahul286:
Support for max_id and min_id is already present in [Telegram API](https://core.telegram.org/method/messages.getHistory), but was not exposed to higher level methods


rahul286:
I will create a different pull request. This one has additional commits unrelated max_id/min_id

PR #1226, opened 2023-02-28
Correct username parsing

KyMaP13:
Its very important when you have more then one username
Usernames

### DIFF
--- a/pyrogram/types/user_and_chats/user.py
+++ b/pyrogram/types/user_and_chats/user.py
@@ -216,6 +216,14 @@ class User(Object, Update):
         if user is None or isinstance(user, raw.types.UserEmpty):
             return None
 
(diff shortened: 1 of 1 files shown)


KyMaP13:
@delivrance hi, it helps to define the username correctly, check this please.

KyMaP13:
I've tested this behavior on pyrogram version 2.0.27 and it works correctly for persons with alternative logins, username is not null, but in new version returned null. My hypothesis is that in new version used new version of mtproto and it gets other data in packets, am i wrong?

lmaxyz:
What `user.username` contains if you have several usernames?

KyMaP13:
> What `user.username` contains if you have several usernames?

Now, its None

PR #1229, opened 2023-03-04
added silent attribute

ivazin:
Added silent attribute to know if the message was sent with sound or not


ivazin:
Please tell me what needs to be amended/how to get this attribute in the main version?

PR #1245, opened 2023-03-29
Fix a bug and add ByteIO to thumb in input medias

meybj:
When you want only edit media of message with edit_message_media, it will remove caption of media because  "" (empty string) have been set as default in input_media_auido (and others which have caption) which will remove caption.

Added ByteIO in some of input medias that support thumbnail.

in guess_mime_type of edit_message_media instead of filename, media was given. when we pass ByteIO to media it will throw error.


mr-st0rm:
hello, can u request changes for your pathc, cause i think it helps me

mr-st0rm (approved):

PR #1261, opened 2023-04-21
Update API schema to Layer 158

null-nick:


null-nick:
Merged commit bb678a4

PR #1313, opened 2023-07-25
Prevent connecting each time to dc when downloading/uploading

KurimuzonAkuma:


ChuckNorrison:
This works great, tested here

https://github.com/KurimuzonAkuma/pyrogram/releases/tag/v2.1.16

Issue #1287, opened 2023-06-20
FloodWait error by auth.ExportAuthorization 


fivepe:
### Checklist

- [X] I am sure the error is coming from Pyrogram's code and not elsewhere
- [X] I have searched in the issue tracker for similar bug reports, including closed ones
- [X] I ran `pip3 install -U https://github.com/pyrogram/pyrogram/archive/master.zip` and reproduced the issue using the latest development version

### Description

I try to download more than 1000 files, and every time I download about 100 files, Waiting for 2856 seconds before continuing (required by "auth.ExportAuthorization") will appear. How to avoid this problem and how to deal with it.

### Steps to reproduce

Login into telegram account
try to download parallel more than 1000 files
There is Waiting for 2856 seconds before continuing (required by "auth.ExportAuthorization") prompt, how to avoid this problem

### Code example

```python
try:
    file = await app.download_media(msgobj, file_name=_filename))
except Exception as E:
    print(f'Exception raised {E}')
```

### Logs

```shell
Telegram says: [420 FLOOD_WAIT_X] - A wait of 1076 seconds is required (caused by "auth.ExportAuthorization")
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/pyrogram/client.py", line 861, in get_file
    exported_auth = await self.invoke(
  File "/usr/local/lib/python3.10/site-packages/pyrogram/methods/advanced/invoke.py", line 79, in invoke
    r = await self.session.invoke(
  File "/usr/local/lib/python3.10/site-packages/pyrogram/session/session.py", line 389, in invoke
    return await self.send(query, timeout=timeout)
  File "/usr/local/lib/python3.10/site-packages/pyrogram/session/session.py", line 357, in send
    RPCError.raise_it(result, type(data))
  File "/usr/local/lib/python3.10/site-packages/pyrogram/errors/rpc_error.py", line 91, in raise_it
    raise getattr(
pyrogram.errors.exceptions.flood_420.FloodWait: Telegram says: [420 FLOOD_WAIT_X] - A wait of 1076 seconds is required (caused by "auth.ExportAuthorization")
```

gr0esch:
I just run in the same problem…

`Telegram says: [420 FLOOD_WAIT_X] - A wait of 142 seconds is required (caused by "auth.ExportAuthorization")
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/pyrogram/client.py", line 880, in get_file
[...]

PR #1348, opened 2023-10-02
Add FloodWait except to media download

jordyamc:
Add FloodWait except to media download to handle this error from the script.

Before this change the exception would only be printed and the script would continue without a way to detect this error.


gr0esch:
Great, this looks like it fixes the bug we've been talking about #1287 and #1306, right? So let's wait for @delivrance to approve it! :)

PR #1402, opened 2024-03-14
Add a method to ForwardMessages class

Erwinex:
Adding a method to ForwardMessages class that works like forward_messages but it won't mention the sender name.

### DIFF
--- a/pyrogram/methods/messages/forward_messages.py
+++ b/pyrogram/methods/messages/forward_messages.py
@@ -33,7 +33,8 @@ class ForwardMessages:
         disable_notification: bool = None,
         schedule_date: datetime = None,
         protect_content: bool = None,
-        hide_author: bool = False
+        hide_author: bool = False,
+        hide_captions: bool = False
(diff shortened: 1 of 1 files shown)


KurimuzonAkuma:
Why not just add drop_author to an existing method... ☠️☠️☠️

Erwinex:
That's definitely a cleaner way thanks I made changes could you check to make sure everything is fine? 🙃

SpEcHiDe:
image

Hi, I think you can expose both `drop_author` and `drop_media_captions` flags as documented in [messages.forwardMessages](https://core.telegram.org/method/messages.forwardMessages)

Erwinex:
Hi, thanks for suggestion I added hide_captions not drop_media_captions becuase I think  hide_author and hide_captions are more user friendly 🫠

hellbowe:
Actually there is `copy_message()` that literally the same as `forward_message()`, but without sender's name

Issue #1314, opened 2023-07-26
PEER_ID_INVALID when send_message to a private group with admin rights


SyberiaK:
### Checklist

- [X] I am sure the error is coming from Pyrogram's code and not elsewhere
- [X] I have searched in the issue tracker for similar bug reports, including closed ones
- [X] I ran `pip3 install -U https://github.com/pyrogram/pyrogram/archive/master.zip` and reproduced the issue using the latest development version

### Description

When my bot gets message, callback data or inline query, it tries to log it into specific private group.
```python
async def log(client: Client, message: Message):
    """The bot sends log to log channel"""

    text = (
        f"✍️ User: <a href=\"tg://user?id={message.from_user.id}\">{message.from_user.first_name}</a>\n"
        f"ID: {message.from_user.id}\n"
        f"Language: {message.from_user.language_code}\n"
        f"Private message: {message.text}"
    )
    await client.send_message(config.LOGCHANNEL, text, disable_notification=True)

@app.on_message()
async def sync_user_data(client: Client, message: Message):
    await log(client, message)
```

Before this date, it did that properly without any error.
However, once I tried to run the bot from another script (to send messages on function call), stopped it and ran this one, the bot started to throw `PeerIdInvalid`:

```python
13:41:50 — 26/07/2023 | pyrogram.dispatcher: Telegram says: [400 PEER_ID_INVALID] - The peer id being used is invalid or not known yet. Make sure you meet the peer before interacting with it 
Traceback (most recent call last):
  File "C:\Users\guard\Desktop\vscode projects\botty\.venv\Lib\site-packages\pyrogram\methods\advanced\resolve_peer.py", line 62, in resolve_peer
    return await self.storage.get_peer_by_id(peer_id)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\guard\Desktop\vscode projects\botty\.venv\Lib\site-packages\pyrogram\storage\sqlite_storage.py", line 142, in get_peer_by_id
    raise KeyError(f"ID not found: {peer_id}")
KeyError: 'ID not found: -914613674'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\guard\Desktop\vscode projects\botty\.venv\Lib\site-packages\pyrogram\methods\advanced\resolve_peer.py", line 85, in resolve_peer
    return await self.storage.get_peer_by_phone_number(peer_id)
[...]

PR #1430, opened 2024-06-09
pyrogram: update MIN_CHANNEL_ID

youesky:
### Contribution Request: Update MIN_CHANNEL_ID in Pyrogram

**Title:** Update MIN_CHANNEL_ID to Fix "Peer id invalid" Error

**Description:**

This contribution request proposes an update to the `MIN_CHANNEL_ID` value in Pyrogram. The current `MIN_CHANNEL_ID` value of `-1002147483647` is causing a "Peer id invalid" error in some scenarios. This update changes the `MIN_CHANNEL_ID` to `-1009999999999` to address this issue.

**Changes:**

- **Old Value:** `MIN_CHANNEL_ID = -1002147483647`
- **New Value:** `MIN_CHANNEL_ID = -1009999999999`

**Reason for Change:**

The existing `MIN_CHANNEL_ID` value is no longer valid in certain cases, leading to the "Peer id invalid" error. By updating the value to `-1009999999999`, we aim to resolve this error and ensure that Pyrogram can handle channel IDs correctly.

**Code Update:**

The change will be made in the relevant Pyrogram file where `MIN_CHANNEL_ID` is defined. 

```python
# Current value
MIN_CHANNEL_ID = -1002147483647

# Updated value
MIN_CHANNEL_ID = -1009999999999
```

**Testing:**

To verify this change, tests will be conducted to ensure that channel IDs are processed correctly without causing the "Peer id invalid" error. Specific test cases will include:

1. Creating a channel with an ID greater than `-1009999999999` and verifying successful handling.
2. Handling channel IDs close to the boundary value to ensure no off-by-one errors or similar issues.
3. General regression testing to ensure no other functionality is impacted.

**Impact:**

This update is expected to fix the "Peer id invalid" error without introducing any breaking changes. All existing functionalities should remain intact, with improved reliability in handling channel IDs.

**Acknowledgements:**

Special thanks to the users and developers who identified and reported the "Peer id invalid" error and contributed to the discussion around the solution.

**Additional Notes:**

Please review and merge this contribution to improve the robustness of Pyrogram in handling Telegram channel IDs. If there are any questions or further suggestions, I am available to discuss them.

Thank you for your time and consideration.

---

*Submitted by:* (ebiza)[telegram.me/ebiza]

PR #1451, opened 2024-09-24
Lower the MIN_CHANNEL_ID to fix the Invalid Peer Id error

ASafarzadeh:
Changed MIN_CHANNEL_ID so we dont face error on other channels.

Previously it was on -1002147483647, that threw errors on some channels like https://t.me/c/2164990312/1 That are legit channels but their chat_id isnt in the range specified.

This fix addresses the issue #1314


SpEcHiDe:
Duplicate of #1430

